PR #8, opened 2022-03-24
Update README to reflect branching model

sdruskat:
Create a `develop` branch (to keep) and document branching model.
:warning: Do not delete branch after merge, rather, set as default branch and lock down `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,8 @@ For more information about the HERMES project, see the [HERMES project website](
 
 1. Clone this repository
 2. If you want to use [`poetry`](https://python-poetry.org), run `poetry shell` and `poetry install`
+
+This project uses 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,3 +14,8 @@ For more information about the HERMES project, see the [HERMES project website](
 
 1. Clone this repository
 2. If you want to use [`poetry`](https://python-poetry.org), run `poetry shell` and `poetry install`
+
+This project uses 
+
+- a development branch (`develop`) to merge developments into, this is the default branch
```
led02, line 20: Maybe hint that actual development should be done in feature branches.

```suggestion
- a development branch (`develop`) to merge developments into, this is the default branch
- actual development is done on feature branches
```

led02 (approved):

PR #15, opened 2022-06-10
Add ADR 0011 about metadata provenance

sdruskat:
This simply adds and ADR draft about metadata provenance.

### DIFF
--- a/docs/adr/0011-record-provenance-of-metadata.md
+++ b/docs/adr/0011-record-provenance-of-metadata.md
@@ -1,0 +1,43 @@
+# Record provenance of metadata
+
+* Status: proposed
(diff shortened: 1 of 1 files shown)


jkelling (approved):
I would vote for that PROV standard: Only using very few of the specified fields is not a strong counter-argument, because standard should be intentionally general, in order to be usable in as many contexts as possible, otherwise it would have a hard time serving as a means of interchange between applications.

jkelling:
I did not, merge, because someone else should also look at this .. and then merge.

sdruskat:
Thanks. Requested a review from @led02, and @poikilotherm's review is still open.

PR #20, opened 2022-06-23
Harvest CFF by using conversion with `cffconvert`

sdruskat:
This PR implements one possible way to implement the path from CFF to HermesContext, using [`cffconvert`](https://github.com/citation-file-format/cff-converter-python) to convert CFF to CodeMeta. Implementation is POC still.

An alternative to this would be to map manually in our package.

Deciding which of the two is more suitable is in scope for the discussion of this PR.


src/hermes/commands/harvest.py
```diff
@@ -29,25 +37,14 @@ def harvest_cff(click_ctx: click.Context, ctx: HermesContext):
         return 1
     else:
```
led02, line 38: `else` is not required as the `if` branch returns. Reduced indentation == reduced complexity (at least if you are a programming metrics tools)

src/hermes/commands/harvest.py
```diff
@@ -29,25 +37,14 @@ def harvest_cff(click_ctx: click.Context, ctx: HermesContext):
...
         cff_dict = load_cff_from_file(cff_file)
-        if not validate(cff_dict):
+        if not validate(cff_file, cff_dict):
             return 1
         else:
```
led02, line 42: See above (reduce complexity, get rid for `else`)

test/hermes_test/test_cli.py
```diff
@@ -102,8 +102,36 @@ def test_haggis_with_post():
     assert not result.exception
 
 
+def test_haggis_with_path():
+    runner = CliRunner()
+    result = runner.invoke(cli.haggis, args=('--path', './'))
+
+    assert not result.exception
```
led02, line 109: A more explicit assertion would be nice. Cannot surely state what, though... maybe `assert result.exception is None`

src/hermes/commands/harvest.py
```diff
@@ -29,25 +37,14 @@ def harvest_cff(click_ctx: click.Context, ctx: HermesContext):
         return 1
```
led02, line 37: Not sure what "the docs" of click say here, but maybe raising an exception is preferrable over return a status code.

(This might also be somehow related to our wrapping `harvest` command... we'll discuss on that maybe)

src/hermes/commands/harvest.py
```diff
@@ -16,6 +17,13 @@
 _CFF_VERSION = '1.2.0'
 
 
+def convert_cff_to_codemeta(cff_file):
```
led02, line 20: Maybe add type annotations...

src/hermes/commands/harvest.py
sdruskat: Good point. Fixed in 2e709b601cae318ec72c93a9b78610a1db3b5cd1.

src/hermes/commands/harvest.py
sdruskat: Thanks again, fixed in  2e709b601cae318ec72c93a9b78610a1db3b5cd1.

src/hermes/commands/harvest.py
sdruskat: Yes, agree that raising an exception is better. Raising `HermesValidationError` in 2e709b601cae318ec72c93a9b78610a1db3b5cd1. Decided that semantically, this is the correct error to raise (instead of creating a new custom one), as per the API docs.

Currently, these aren't caught anywhere, so I agree that more discussion is necessary. (e.g., try/except in `harvest.py` or elsewhere?)

test/hermes_test/test_cli.py
sdruskat: Not sure how this works in Python, but isn't `assert result.exception is None` the same as `assert not result.exception`?

src/hermes/commands/harvest.py
sdruskat: Done in f6190b1882d9cba79b692ab59888471fd16c50c8. Also fixed some errors that mypy reported.

test/hermes_test/test_cli.py
led02: Technically: no.

`assert not result.exception` is coerced to `assert not bool(result.exception)` which might have some unforseen effects. E.g., if you have something like `throw ''` (i.e., you throw an empty string instead of a proper exception, which is allowed) you'd end up with `not result.exception` evaluating to `True` while `result.exception is not None` evaluating to `False`.

Also, some might think that the proposed version would make a better PyTest backtrace (but this is speculative).

led02: Re-thinking this (and looking at `git blame`) the `not result.exception` looks really great... 😇  However, more explicit might be still be a good idea (but I won't deny my approval for this).

poikilotherm:
@sdruskat: @led02 and me had a nice pair programming session on this stuff today and got it polished up a bit :wink: 

We left some TODOs to look at for later though.

PR #21, opened 2022-06-23
Harvesting skeleton

### LINKED ISSUE #18: HermesContext should accept bulk data
Hi @led02, there are converters to CodeMeta for some metadata sources out there (e.g., `cffconvert` for CFF). Hence, in the harvesting pipeline, these could be reused to create CodeMeta JSON-LD from the source.

Then, it would be really nice if the `HermesContext` would accept the whole CodeMeta JSON-LD in bulk, e.g., `ctx.bulk_update(json.loads(codemeta_str), src=source)`.

What do you think?

led02:
Fixes #18

However, there is still room for discussion of the interface. Especially, if the `data` dictionary should be *unrolled* (but I'd say yes for reasons)...

### DIFF
--- a/src/hermes/model/context.py
+++ b/src/hermes/model/context.py
@@ -1,9 +1,166 @@
+import traceback
+import json
+import logging
+import typing as t
+
+from pathlib import Path
+from importlib.metadata import EntryPoint
+
+from hermes.model.errors import HermesValidationError
+
+
+_log = logging.getLogger(__name__)
+
+
 class HermesContext:
-    def __init__(self):
+    """
+    The HermesContext stores the metadata for a certain project.
+
+    As there are different views of the metadata in the different stages,
+    some stages use a special subclass of this context:
+
+    - The *harvest* stages uses :class:`HermesHarvestContext`.
+    """
+
+    def __init__(self, project_dir: t.Optional[Path] = None):
+        """
+        Create a new context for the given project dir.
+
+        :param project_dir: The root directory of the project.
+                            If nothing is given, the current working directory is used.
+        """
+
+        #: Base dir for the hermes metadata cache (default is `.hermes` in the project root).
+        self.hermes_dir = Path(project_dir or '.') / '.hermes'
+
+        self._caches = {}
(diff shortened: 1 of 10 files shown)


src/hermes/model/context.py
```diff
@@ -0,0 +1,215 @@
...
+            for key, value in _value.items():
+                self._update_key_from(f'{_key}.{key}', _value, **kwargs)
+            else:
+                self.update(_key, _value, **kwargs)
+
+    def update_from(self, data: t.Dict[str, t.Any], **kwargs: t.Any):
+        """
+        Bulk-update multiple values.
+
+        If the value for a certain key is again a dictionary, the key will be expanded. I.e.
```
led02, line 188: In fact, we did not really discuss this behaviour but I'd think for the harvesting it might be a good idea...

src/hermes/model/context.py
```diff
@@ -0,0 +1,215 @@
...
+            if entry[1] == kwargs:
+                self._log.debug("Update %s: %s -> %s (%s)", _key, entry[0], _value, entry[1])
+                entry[0] = _value
+                break
+        else:
+            self._data[_key].append([_value, kwargs])
+
+    def _update_key_from(self, _key: str, _value: t.Any, **kwargs):
+        if isinstance(_value, dict):
+            for key, value in _value.items():
```
led02, line 179: Should we add a similar behaviour for arrays?

src/hermes/model/context.py
sdruskat: Agree, if you mean that the extra values are also recorded?

src/hermes/model/context.py
sdruskat: I think dictionary is fine for now.

src/hermes/model/context.py
sdruskat: There's some explanation missing after "I.e." ;)

sdruskat (approved):
@led02 LGTM, thanks!
Feel free to merge 🌮 !

PR #23, opened 2022-07-06
Create ADR about data model (high level)

led02:


sdruskat (approved):
Good start for a proposed ADR. Thanks! 🍰

PR #24, opened 2022-07-11
7 - add initial docs configuration

### LINKED ISSUE #7: Set up documentation
We need the following documentation:

- User docs (plain text)
    - End users
    - Insitutional users (e.g., users that set up, extend, configure the workflow for their own instances of target repos)
- Dev docs
    - docstrings (rendered into package docs)
    - plain text docs
    - ADRs
- Contribution documentation (cf. #5)

The question is:

- Should we provide all docs types in one "document", alternatively, use dedicated tooling for each type (or a subset)

poikilotherm:
Based on what we did for https://project.software-metadata.pub, this is an initial setup of our workflow docs.

Closes #7

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -8,7 +8,7 @@ python-versions = "*"
 
 [[package]]
 name = "atomicwrites"
-version = "1.4.0"
+version = "1.4.1"
 description = "Atomic file writes."
 category = "dev"
 optional = false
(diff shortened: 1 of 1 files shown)


led02:
Two remarks:

- Moving the ADRs breaks usage of https://adr.github.io/adr-manager
- Why did you introduce a new `requirements.txt` instead of using Poetry tooling?

Except for that, the documentation config looks good.

led02:
> Can we separate the docs deps there? Also, `requirements.txt` is properly supported at RTD, not sure about Poetry.

Poetry allow to have "developer-dependencies" at least. However, the point with RTD seems valid for me....

led02 (approved):
LFTM 🌮

PR #25, opened 2022-09-01
Add simple git author harvester

jkelling:
This patch adds a harvester that collects some authorship data from git history.

I am not sure how to put the *set* of authors and their (potential) *set* of email addresses into the `HermesHarvestContext`. Using the available array pattern does not work, because order of authors and email addresses may change between runs if the history contains new commits or when run on a different branch (likewise when sorting by name, if a new author joined).

In this version, I solved this by adding the author name as a tag value, rather than as value, because tags act like a set. This leads to a weird pattern for the email, where the address is also a tag value, to get a set, but also a value (same as would happen with an `author.name` key).

### DIFF
--- a/src/hermes/commands/harvest/git.py
+++ b/src/hermes/commands/harvest/git.py
@@ -16,18 +16,18 @@ from hermes.model.errors import HermesValidationError
 # TODO: can and should we get this somehow?
 SHELL_ENCODING = 'utf-8'
 
-class AuthorData:
-    def __init__(self, line: t.List):
-        self.name = line[0]
-        self.email = set((line[1],))
-        self.tFirst = line[2]
+class ConributorData:
+    def __init__(self, d: t.List):
+        self.name = d[0]
+        self.email = set((d[1],))
+        self.tFirst = d[2]
         self.tLast = self.tFirst
 
-    def update(self, line: t.List):
-        assert(self.name == line[0])
+    def update(self, d: t.List):
+        assert(self.name == d[0])
 
-        self.email.add(line[1])
-        t = line[2]
+        self.email.add(d[1])
+        t = d[2]
         if t < self.tFirst:
             self.tFirst = t
         elif t > self.tFirst:
@@ -58,27 +58,36 @@ def harvest_git(click_ctx: click.Context, ctx: HermesHarvestContext):
 
     # Get history of currently checked-out branch
     authors = {}
-    p = subprocess.run([gitExe, "log", "--pretty=%an_%ae_%ad", "--date=unix"], capture_output=True)
+    committers = {}
+    p = subprocess.run([gitExe, "log", "--pretty=%an_%ae_%at_%cn_%ce_%ct"], capture_output=True)
     if p.returncode:
         raise RuntimeError("`git log` command failed with code {}: '{}'!".format(p.returncode, p.stderr.decode(SHELL_ENCODING)))
 
     log = p.stdout.decode(SHELL_ENCODING).split('\n')
     for l in log:
         d = l.split('_')
-        if len(d) != 3:
+        if len(d) != 6:
             continue
         try:
             d[2] = int(d[2])
         except ValueError:
             continue
 
-        if d[0] in authors:
-            authors[d[0]].update(d)
-        else:
-            authors[d[0]] = AuthorData(d)
+        _updateContributor(authors, d[0:3])
+        _updateContributor(committers, d[3:7])
     
-    for a in authors.values():
-        ctx.update("author.since", a.tFirst, name=a.name, branch=gitBranch)
-        ctx.update("author.until", a.tLast, name=a.name, branch=gitBranch)
+    _ctxUpdateContributors(ctx, authors, "author", branch=gitBranch)
+    _ctxUpdateContributors(ctx, committers, "committer", branch=gitBranch)
+
+def _updateContributor(contributors: t.Dict, d: t.List):
+    if d[0] in contributors:
+        contributors[d[0]].update(d[0:3])
+    else:
+        contributors[d[0]] = ConributorData(d[0:3])
+
+def _ctxUpdateContributors(ctx: HermesHarvestContext, contributors: t.Dict, kind: str, **kwargs):
+    for a in contributors.values():
+        ctx.update(f"{kind}.since", a.tFirst, name=a.name, **kwargs)
+        ctx.update(f"{kind}.until", a.tLast, name=a.name, **kwargs)
         for e in a.email:
-            ctx.update("author.email", e, name=a.name, branch=gitBranch, email=e)
+            ctx.update(f"{kind}.email", e, name=a.name, email=e, **kwargs)
--- a/src/hermes/commands/harvest/git.py
+++ b/src/hermes/commands/harvest/git.py
@@ -16,22 +16,24 @@ from hermes.model.errors import HermesValidationError
 # TODO: can and should we get this somehow?
 SHELL_ENCODING = 'utf-8'
 
+
 class ConributorData:
     def __init__(self, d: t.List):
         self.name = d[0]
         self.email = set((d[1],))
-        self.tFirst = d[2]
-        self.tLast = self.tFirst
+        self.t_first= d[2]
+        self.t_last= self.t_first
 
     def update(self, d: t.List):
         assert(self.name == d[0])
 
         self.email.add(d[1])
         t = d[2]
-        if t < self.tFirst:
-            self.tFirst = t
-        elif t > self.tFirst:
-            self.tLast = t
+        if t < self.t_first:
+            self.t_first = t
+        elif t > self.t_first:
+            self.t_last = t
+
 
 def harvest_git(click_ctx: click.Context, ctx: HermesHarvestContext):
     """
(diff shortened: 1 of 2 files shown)


src/hermes/commands/harvest/git.py
```diff
@@ -0,0 +1,84 @@
...
+        if d[0] in authors:
+            authors[d[0]].update(d)
+        else:
+            authors[d[0]] = AuthorData(d)
+    
+    for a in authors.values():
+        ctx.update("author.since", a.tFirst, name=a.name, branch=gitBranch)
+        ctx.update("author.until", a.tLast, name=a.name, branch=gitBranch)
+        for e in a.email:
+            ctx.update("author.email", e, name=a.name, branch=gitBranch, email=e)
```
sdruskat, line 84: Based on software citation principles, the semantics are off for updating values with the `author` prefix. This is because the only thing we really know about these entities (could include bots) is that they have pushed commits. Authors can only be defined by the software project itself, so we cannot rely on git here. Perhaps update the context with the prefix `committer` instead? 

It would probably also be more "correct" to rename the vars.

src/hermes/commands/harvest/git.py
```diff
@@ -0,0 +1,84 @@
...
+import shutil
+
+from hermes.model.context import HermesHarvestContext
+from hermes.model.errors import HermesValidationError
+
+# TODO: can and should we get this somehow?
+SHELL_ENCODING = 'utf-8'
+
+class AuthorData:
+    def __init__(self, line):
```
sdruskat, line 20:
```suggestion
    def __init__(self, line: t.List):
```

Add type hint.

sdruskat (commented):
Thanks, @jkelling, for this PR! As for the `set` issue, I'd guess we need to type the model a bit stronger, but would wait for @led02's input on this.

src/hermes/commands/harvest/git.py
jkelling: git has separate concepts for "author" and "committer". So, if someone commits someone else's code, and use git properly, the info will be correct. Committers could be harvested additionally.

jkelling: Incidentally, see https://github.com/hermes-hmc/workflow/pull/25/commits/acfdf826c27d6d550ee4bb8f92dcf8ffd489afd0 .

sdruskat: > git has separate concepts for "author" and "committer". So, if someone commits someone else's code, and use git properly, the info will be correct. Committers could be harvested additionally.

Ah, I see. In this case, we should find some other term to clarify that these are *authoritative authors* as defined by the humans in the project, or what git thinks is an author. Perhaps "contributor" is better, as we definitely have contributions here.

led02: contributor?

jkelling: The harvested data in the cache is pre-curation anyway. There must be a curation step in which the user decides how to use the harvested data. Consider also inconsistent spelling of names or multiple listed email addresses. For the cache, it may be better stick to the terminology used in the data source to avoid confusion.

src/hermes/commands/harvest/git.py
```diff
@@ -0,0 +1,93 @@
...
+        except ValueError:
+            continue
+
+        _updateContributor(authors, d[0:3])
+        _updateContributor(committers, d[3:7])
+    
+    _ctxUpdateContributors(ctx, authors, "author", branch=gitBranch)
+    _ctxUpdateContributors(ctx, committers, "committer", branch=gitBranch)
+
+def _updateContributor(contributors: t.Dict, d: t.List):
```
led02, line 82:
```suggestion
        _updateContributor(authors, d[0:3])
        _updateContributor(committers, d[3:7])
    
    _ctx_update_contributors(ctx, authors, "author", branch=git_branch)
    _ctx_update_contributors(ctx, committers, "committer", branch=git_branch)

def _update_contributor(contributors: t.Dict, d: t.List):
```

src/hermes/commands/harvest/git.py
```diff
@@ -0,0 +1,93 @@
...
+    if d[0] in contributors:
+        contributors[d[0]].update(d[0:3])
+    else:
+        contributors[d[0]] = ConributorData(d[0:3])
+
+def _ctxUpdateContributors(ctx: HermesHarvestContext, contributors: t.Dict, kind: str, **kwargs):
+    for a in contributors.values():
+        ctx.update(f"{kind}.since", a.tFirst, name=a.name, **kwargs)
+        ctx.update(f"{kind}.until", a.tLast, name=a.name, **kwargs)
```
led02, line 91:
```suggestion


def _ctx_update_contributors(ctx: HermesHarvestContext, contributors: t.Dict, kind: str, **kwargs):
    for a in contributors.values():
        ctx.update(f"{kind}.since", a.t_first, name=a.name, **kwargs)
        ctx.update(f"{kind}.until", a.t_last, name=a.name, **kwargs)
```

led02 (changes requested):
PEP-ified

jkelling:
> PEP-ified

Thanks... is there a style-file included in the repo ... and a hint on how to run a fixer?

src/hermes/commands/harvest/git.py
led02: Okay, this seems like a good argument ... even though this could be done automagically by the associated "Pre-Processor" and not during manual curation.  But this needs to be implemented yet and I think it's a beautiful example to investigate into implementation strategies for this.
So (except for the PEP-8-flaws) I like your contribution and am happy to approve it :)

led02 (approved):

led02:
> Thanks... is there a style-file included in the repo ... and a hint on how to run a fixer?

Well, good question actually.

- In ADR 0009 we define that we want to adhere "to Python style" which is in fact a bit short. However, there is [PEP8](https://pep8.org/) which is widely adopted as Python style. Maybe we can clearify this for those not too deep into Python yet.
- Style checkers exist but are not yet integrated into the workflow. Most common are [pylint](https://pylint.pycqa.org/en/latest/) or [black](https://github.com/psf/black).
- I personally do not like autoformatters, hence I cannot give you a hint which works best. But if you search for "PEP8 formatter" or something alike, you will find a range of possible tools. I personally use PyCharm as my IDE which automatically checks for PEP8 and provides quick fixes.

sdruskat:
> > Thanks... is there a style-file included in the repo ... and a hint on how to run a fixer?
> 
> Well, good question actually.
> 
>     * In ADR 0009 we define that we want to adhere "to Python style" which is in fact a bit short. However, there is [PEP8](https://pep8.org/) which is widely adopted as Python style. Maybe we can clearify this for those not too deep into Python yet.
> 
>     * Style checkers exist but are not yet integrated into the workflow. Most common are [pylint](https://pylint.pycqa.org/en/latest/) or [black](https://github.com/psf/black).
> 
>     * I personally do not like autoformatters, hence I cannot give you a hint which works best. But if you search for "PEP8 formatter" or something alike, you will find a range of possible tools. I personally use PyCharm as my IDE which automatically checks for PEP8 and provides quick fixes.

#42 introduces a CI workflow using `flake8`, you can use it locally as well: 

```bash
flake8 ./test/ ./src/ --count --select=E9,F63,F7,F82 --show-source --statistics
flake8 ./test/ ./src/ --count --max-complexity=10 --max-line-length=120 --statistics
```

@jkelling fell free to merge (as per https://github.com/hermes-hmc/workflow/pull/41/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R74-R76)

jkelling:
How would this conflict? This PR adds a harvester for stage 1, while #26 requests a stage 2, where the harvested data is converted and, I guess, reduced and prepared for curation...

led02:
> How would this conflict? This PR adds a harvester for stage 1, while #26 requests a stage 2, where the harvested data is converted and, I guess, reduced and prepared for curation...

In fact, I merged your branch into #40 to have data that can be merged.

Also, the branch ended up with lots of changes to different parts (context access, logging, …). Hence, adapted the code accordingly.

jkelling:
> > How would this conflict? This PR adds a harvester for stage 1, while #26 requests a stage 2, where the harvested data is converted and, I guess, reduced and prepared for curation...
> 
> In fact, I merged your branch into #40 to have data that can be merged.

I do not like merging .. , and you also did not,: It looks like you based the branch in #40 in this branch. So this PR could be merged and then you rebase your branch onto master. That would not cause conflicts with a fast-forward merge, not sure if a merge commit would cause trouble.

Side note: Using merge commits, kinda makes this harvester useless, because all the authorship information of the merged branch gets lost, only the author of the merge commit remains on the branch merged into... .

led02:
> I do not like merging .. , and you also did not,: It looks like you based the branch in #40 in this branch. So this PR could be merged and then you rebase your branch onto master. That would not cause conflicts with a fast-forward merge, not sure if a merge commit would cause trouble.

Well, in fact I branched `26-implement-processing-step` off `develop`, did one commit (https://github.com/hermes-hmc/workflow/commit/4a61db3a43a52cfe8c7674535d39adb744a24526) and then merge `githarvester`. You are right that a *rebase* should not lead to bigger problems.

As to "Merge Commit" vs. "Rebase" this is a pretty philosophical question and has to do a lot with preferences... (I infact prefer to have merges.) We decided not to prefer one of those processes (yet). The main point is however: *There were lots of changes to the githarvester after the refactoring for the processing step. Hence, your code wouldn't probably work anymore with the current architecture.*

However, if you prefer to have this branch merged and then #40 rebased, it's okay. My main focus is to get #40 merged as soon as possible so that we can continue with development... Maybe you could have a look into that pull request. too.

led02 (approved):
Okay to merge and then rebase the branch 26-implement-processing-step (see comments)

jkelling:
> Okay to merge and then rebase the branch 26-implement-processing-step (see comments)

There two lines were missed when PEP8-ifying, leading to undefined identifiers. These would need to be fixed before merging, which would cause a conflict when rebasing #40, because these lines were removed.

I will close this PR, to have the feature merged with #40, then

PR #43, opened 2022-09-22
Fix 4: license project and provide copyright info

sdruskat:
This PR fixes #4.

- Add the [REUSE GitHub Action](https://github.com/marketplace/actions/reuse-compliance-check)
- Add license and copyright information to all files

# Review checklist

- [ ] Check **carefully** all license and copyright information
- [ ] Check that the REUSE Specs GitHub Action passes
- [ ] Check that `LICENSE.md` contains useful content

PR #78, opened 2023-01-10
Fix full term for ADR in docs

sdruskat:
The docs still call ADRs _Architecture Design Records_ instead of the correct _Architectural Decision Records_. This PR fixes that, and also adds a missing extension to the docs `requirements.txt` (that we need because Sphinx won't play nice with poetry yet).

To test, do

```bash
# On develop to see the error
cd docs
python3.11 -m venv venv
. venv/bin/activate
make deps && make livehtml

# Error:
# Running Sphinx v4.5.0
# loading translations [en]... done
#
# Extension error:
# Could not import extension autoapi.extension (exception: No module named 'autoapi')

git switch feature/fix-terminology-in-docs

rm -rf venv/
python3.11 -m venv venv
. venv/bin/activate

make deps && make livehtml
# Installs and builds without errors
```

### DIFF
--- a/docs/adr/index.md
+++ b/docs/adr/index.md
@@ -4,7 +4,7 @@ SPDX-FileCopyrightText: 2022 German Aerospace Center (DLR), Forschungszentrum JÃ
 SPDX-License-Identifier: CC-BY-SA-4.0
 -->
 
-# Architecture Design Records
+# Architectural Decision Records
 
 ```{toctree}
 :glob:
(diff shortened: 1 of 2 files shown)


poikilotherm:
For some reason, #43 re-introduced the old doc build files. Removing them here, as we do use `poetry` (see README).

poikilotherm (approved):
LGTM

PR #92, opened 2023-01-11
Rename post to postprocess in various places

zyzzyxdonta:
This pull request renames "post" to "postprocess" in various places. This helps distinguishing it from the verb "to post" which is not what it does. Also, it makes sure that the noun "post-processing" is spelled with a hyphen, and the verb is spelled without a hyphen.


sdruskat (approved):
LGTM 🌮, thanks @zyzzyxdonta!

PR #104, opened 2023-01-12
Add user feeback when harvesting fails due to (lack of) data sources

### LINKED ISSUE #30: Add user feedback when no data sources found
I ran `haggis` in a random directory with no data sources, and the only output was:
```
% haggis 
Metadata harvesting
Metadata processing
```
It would be good to add some user feedback on the command line to say that no data sources were found.

jkelling:
Closes #30

### DIFF
--- a/src/hermes/model/context.py
+++ b/src/hermes/model/context.py
@@ -173,6 +173,9 @@ class HermesHarvestContext(HermesContext):
         if exc_type is not None and issubclass(exc_type, HermesValidationError):
             exc = traceback.TracebackException(exc_type, exc_val, exc_tb)
             self._base.error(self._ep, exc)
+            self._log.warning("{}: {}".format(
+                exc_type.__name__,
+                ' '.join(str(x) for x in exc_val.args)))
             return True
 
     def update(self, _key: str, _value: t.Any, **kwargs: t.Any):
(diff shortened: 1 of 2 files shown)


src/hermes/model/context.py
```diff
@@ -173,6 +173,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is not None and issubclass(exc_type, HermesValidationError):
             exc = traceback.TracebackException(exc_type, exc_val, exc_tb)
             self._base.error(self._ep, exc)
+            self._log.warning("{}: {}".format(
+                exc_type.__name__,
+                ' '.join(str(x) for x in exc_val.args)))
```
led02, line 178:
```suggestion
            self._log.warning("%s: %s",
                exc_type,
                ' '.join(map(str, exc_val.args)))
```

While your change is totally fine, this proposal makes it more "pythonic" and slightly improves performance:

- Using `self._log.warning("%s: %s", ...)` leads to evaluation of the format string only in cases where the log message is actually rendered (i.e., if it is not filtered out by the logger or handler)
- Similar using `map(str, ...)` is only evaluated once the content is actually iterated. However, this is in fact already done when `' '.join(...)` is called and thus does not improve performance a lot here...

led02 (approved):
Looks good, thank you.

Issue #46, opened 2022-09-28
Define and implement configuration strategy


sdruskat:
Users will need to be able to configure the workflow quite extensively, e.g.,:

- Which plugins should be run?
- How to map alternative email addresses?
- ...

We need to define a confiuration strategy that includes a number of things:

- Should there be a one-size-fits-all config file?
- Should there be additionally a way to configure as params?
- Should we wrap, or simply reuse, native solutions such as `.mailmap` (standard git email mapping config)?

This strategy will have to satisfy a good UX! 

E.g., as a HERMES user, I may not know about `.mailmap` or similar, and may not want to be bothered with another config file, especially when I'm already working with one thing (the HERMES config).

led02:
I would prefer "native solutions" where possible, just because they are then picked up by other tools, too. (No duplication required.)

However, I also see a problem because a clean approach would mean to also use "native solutions" in the other direction. I.e., HERMES should automatically identify what kind of project it is applied to and detect relevant configuration files (e.g., `pyproject.tom`, `pom.xml`, `Cargo.toml`, ...). This would however be just too much different possibilities to handle.

Hence, my proposal is:

- Have a unified, independent `hermes.toml`
- Support the user in using "native solutions" (like `.mailmap`, which is not read by Hermes tool at all but leads to expected results)
- Keep in mind for later that using the "build configuration" file might be an extensible solution.

sdruskat:
I agree with a unified config for hermes.

- **question:** Why `.toml` rather than `.yml`?
- **question:** If we support the user in using native solutions, but don't pick it up later, isn't this either out-of-scope, or wasted time?
    - **suggestion:** Can we support both approaches at the same time, leaving it up to the user, which configuration path to choose? I.e., implement support for `.mailmap`-style mail mapping in `hermes.*ml` (and prefer that over native solutions, as it is more expressive), but fall back to native solutions if it isn't configured?

sdruskat:
See also #39 for a more implementation-centric discussion.

sdruskat:
Decisions:

- We wil provide a Python script tied to a `hermes` command that lets users update configuration files
- We'll use `hermes.conf` in TOML
- Native solutions are not yet systematically supported, although they're useful to support for users. Tracking this in #102.

PR #106, opened 2023-01-12
Make pyproject.toml optional by adding default logging config

### LINKED ISSUE #100: `hermes` needs pyproject.toml for logging config

led02:
Fixes #100

### DIFF
--- a/src/hermes/config.py
+++ b/src/hermes/config.py
@@ -14,9 +14,9 @@ _config = {
         'version': 1,
 
         'formatters': {
-            'plain': { 'format': "%(message)s" },
-            'logfile': { 'format': "%(created)16f:%(name)20s:%(levelname)10s | %(message)s" },
-            'auditlog': { 'format': "%(asctime)s %(name)-20s  %(message)s" },
+            'plain': {'format': "%(message)s"},
+            'logfile': {'format': "%(created)16f:%(name)20s:%(levelname)10s | %(message)s"},
+            'auditlog': {'format': "%(asctime)s %(name)-20s  %(message)s"},
         },
 
         'handlers': {
--- a/src/hermes/config.py
+++ b/src/hermes/config.py
@@ -53,14 +53,14 @@ _config = {
 
 
 def configure():
-    if _config:
+    if 'hermes' in _config:
         return
 
     # Load configuration if not present
     try:
         with open('pyproject.toml', 'r') as config_file:
             config_toml = toml.load(config_file)
-            hermes_config = config_toml['tool']['hermes']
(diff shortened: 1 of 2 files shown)


src/hermes/config.py
```diff
@@ -9,19 +9,63 @@
...
         return
 
     # Load configuration if not present
-    with open('pyproject.toml', 'r') as config_file:
-        config_toml = toml.load(config_file)
-        hermes_config = config_toml['tool']['hermes']
-        _config['hermes'] = hermes_config
-        _config['logging'] = hermes_config['logging']
+    try:
+        with open('pyproject.toml', 'r') as config_file:
```
jkelling, line 61: Does this mean, that if there is a hermes config (maybe from some `hermes.toml`) already loaded, a `pyproject.toml` will be ignored, making it mandatory to have the loggin config in the `hemeres.toml` even if one has a`pyproject.toml`?

zyzzyxdonta (approved):
Looks good; defaults work, configuration via pyproject.toml still works as well 👍🏻

src/hermes/config.py
```diff
@@ -9,19 +9,63 @@
...
+            config_toml = toml.load(config_file)
+            hermes_config = config_toml['tool'].get('hermes', {})
+            _config['hermes'] = hermes_config
+            _config['logging'] = hermes_config.get('logging', _config['logging'])
```
jkelling, line 65: If there is no `"logging" in _config`, should setting the second argument of `get()` not throw an exception? Would this not make the get from `hermes_config` ineffective?

led02: The second arg is the default logging configuration as explicited in lines 13 through 51. Using the way here means that a `logging` section in the hermes config would override the defaults completely.

In fact there is an issue here as hermes cli expects certain loggers to be configured to work... maybe a 'merge' of the default config would be a good idea. I'll take that into account when working on #46.

src/hermes/config.py
led02: Yes, there is only one configuration file that should define the behaviour of hermes cli.

jkelling: If it is possible to draw the logging config from `pyproject.toml` if there is no `hernes.toml`, then it would be helpful to be able to do it in either case: maybe someone has the logging config for their project in the `pyproject.toml` and does not want to duplicate the same config in `hermes.toml`. It should also not override an existing config completetly, but only update keys which are set in the new config. THis mechanisn is required anyway to handle defaults.

jkelling:
Merged. @led02 I suggest to rebase #46 , because this is more of a temporary fix, to be done properly in #46 .

PR #107, opened 2023-01-12
Fix #65: Add git authors and committers as contributors

### LINKED ISSUE #65: Split authors and contributors
- **Requires:** #52 

Authors and contributors of software are different roles and should end up in different CodeMeta fields.

E.g., 
- person1 is in CFF authors AND git contributors: person is author only
- person1 is in CFF authors AND NOT git contributors: person is author only
- person1 is NOT in CFF authors BUT in git contributors: person is contributor only

This must be implemented as processing step.

sdruskat:
This PR fixes #65.

For a discussion why this change is necessary, see the issue.

Changes:

- Add all git authors and git committers as contributors.
- Decouple git harvesting and preprocessing from authors, as semantically, we can never assert git contributors to be authors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ hermes = "hermes.cli:main"
 
 [tool.poetry.plugins."hermes.preprocess"]
 000_cff = "hermes.commands.process.cff:add_name"
-020_git = "hermes.commands.process.git:flag_authors"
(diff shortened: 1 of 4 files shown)


poikilotherm (approved):
Thanks for the walk through, looks good to me.

PR #108, opened 2023-01-12
Add command clean

### LINKED ISSUE #81: Add a build clean (e.g., `hermes clean`) that deletes caches
This is important to avoid confusing outputs where conflicts have been resolved but older caches are used for processing.

This is a command or flag.

Things to consider:

- When to automatically overwrite/invalidate caches (e.g., based on timestamp)

jkelling:
Closes #81


src/hermes/model/context.py
```diff
@@ -119,6 +120,14 @@ def error(self, ep: EntryPoint, error: Exception):
 
         self._errors.append((ep, error))
 
+    def purge_caches(self, *path: str) -> Path:
```
led02, line 123: path does not seem to be used, although I think it's good to make this method reusable.

zyzzyxdonta, line 123:
```suggestion
    def purge_caches(self, *path: List[str]) -> None:
```

This method doesn't return anything. Also `path` should probably be a list of str or a list of pathlib.Path.

The path parameter is unused. If it doesn't make sense to implement this now but is supposed to be implemented later, maybe add:

```python
if path:
    raise NotImplementedError()
```

src/hermes/cli.py
```diff
@@ -157,3 +157,4 @@ def main(ctx: click.Context, *args, **kwargs) -> None:
 main.add_command(workflow.process)
 main.add_command(workflow.deposit)
 main.add_command(workflow.postprocess)
+main.add_command(workflow.clean)
```
led02, line 160: This way, clean is *always* run when `hermes` is called without an explicit sub-command. I'm not sure this is what we really want...

zyzzyxdonta (changes requested):

src/hermes/cli.py
led02: My proposal:
- make `clean` the first command
- add a `click.option("--clean", is_flag=True, default=False)` to the main. This way, clean is automatically run as first command as if you call `hermes --clean`.

led02 (changes requested):
See notes. We should discuss *when* cleanup is made. Also, the purge_caches function can be improved.

src/hermes/model/context.py
zyzzyxdonta: OK, the List thing is probably wrong, sorry 😄

src/hermes/cli.py
jkelling: To the first point:  What does the `invoke_without_command` parameter in the decorator mean?
https://github.com/hermes-hmc/workflow/pull/108/files#diff-9c665150c60db6c7a4564d46b1beb793fc2ec2bb2df11aeae0e132c0e8c5bf99R111
I thought setting it to `False` would make it not run, but this lead to some error. I the changed it back and forgot about it, sorry.

jkelling: To the second point: I think we do not want to run it automatically before other commands. We also do not want to run it only when other commands are run either, it should be available as a standalone action which one can run, e.g., after completing a publication process.

src/hermes/model/context.py
jkelling: I guess the path should be remove here. If we want to selectively remove caches, the function should be called `purge_cache` or `remove_cache`.

jkelling: The `path` argument is obsolete, because the path of the hermes project is provided by the context. I first thought, I need to pass the path to the function an the forgot to remove it once I figured-out the context.

jkelling: This functionality would be akin  to `git`'s `-C` option and should be implemented like it, which means it would set the hermes context's path. See https://github.com/hermes-hmc/workflow/pull/108#discussion_r1069185870 .

src/hermes/cli.py
jkelling: Ok, I think I see not how `click` works. I implemented your suggestion.

jkelling:
Fixed and rebase onto develop (to allow me to run flak8 locally :grin: )

src/hermes/model/context.py
```diff
@@ -119,6 +120,14 @@ def error(self, ep: EntryPoint, error: Exception):
 
         self._errors.append((ep, error))
 
+    def purge_caches(self) -> Path:
```
zyzzyxdonta, line 123:
```suggestion
    def purge_caches(self) -> None:
```

This still doesn't return anything

zyzzyxdonta (changes requested):

zyzzyxdonta (approved):
Looks good 👍🏻

led02 (approved):

PR #115, opened 2023-01-13
Add test for git harvester

### LINKED ISSUE #113: Test git harvesting
Needs monkeypatching `subprocess.run`ning the git log --pretty call and matching the harvesting context contents.

sdruskat:
- Fix #113
- Assert the JSONized harvest context after running the git harvester

### DIFF
--- a/test/hermes_test/commands/harvest/test_git.py
+++ b/test/hermes_test/commands/harvest/test_git.py
@@ -9,7 +9,7 @@ from importlib.metadata import EntryPoint
 
 import pytest
 from pytest import FixtureRequest
-from unittest.mock import MagicMock, patch, Mock
+from unittest.mock import patch, Mock
 
 import hermes.commands.harvest.git as harvest
 from hermes.model.context import HermesContext, HermesHarvestContext
@@ -104,7 +104,7 @@ def mock_click_ctx():
 @patch("hermes.commands.harvest.git.subprocess.run")
 def test_harvest_git(mock_run, mock_click_ctx, harvest_ctx, git_log_output, codemeta):
     # Mocking the output of the subprocess call to git log to retrieve author and committer info
-    mock_stdout = MagicMock()
+    mock_stdout = Mock()
     mock_stdout.configure_mock(
         **{
             "stdout.decode.return_value": git_log_output,
--- a/test/hermes_test/commands/harvest/test_git.py
+++ b/test/hermes_test/commands/harvest/test_git.py
@@ -1,0 +1,125 @@
+# SPDX-FileCopyrightText: 2022 German Aerospace Center (DLR)
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# SPDX-FileContributor: Stephan Druskat
+
+import json
+from importlib.metadata import EntryPoint
+
+import pytest
+from pytest import FixtureRequest
+from unittest.mock import MagicMock, patch, Mock
+
+import hermes.commands.harvest.git as harvest
+from hermes.model.context import HermesContext, HermesHarvestContext
+
+
+@pytest.fixture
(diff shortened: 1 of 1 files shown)


test/hermes_test/commands/harvest/test_git.py
```diff
@@ -0,0 +1,125 @@
...
+def mock_click_ctx():
+    ctx = Mock()
+    ctx.parent = Mock(params={'path': '.'})
+    return ctx
+
+
+@patch("hermes.commands.harvest.git.subprocess.run")
+def test_harvest_git(mock_run, mock_click_ctx, harvest_ctx, git_log_output, codemeta):
+    # Mocking the output of the subprocess call to git log to retrieve author and committer info
+    mock_stdout = MagicMock()
```
zyzzyxdonta, line 107: Is `MagicMock` necessary here? It seems to work just fine with `Mock`.

test/hermes_test/commands/harvest/test_git.py
```diff
@@ -0,0 +1,125 @@
...
+            "returncode": 0
+        }
+    )
+
+    # Mocking the output of the subprocess call to git rev-parse to retrieve branch id
+    mock_get_branch = Mock()
+    mock_get_branch.returncode = 0
+
+    # Define outputs from "running" the subprocess mock
+    mock_run.side_effect = [mock_get_branch, mock_stdout]
```
jkelling, line 120: `subprocess.run` is called two times. The mock only provides valid output for the second call. The first call
https://github.com/hermes-hmc/workflow/blob/f02ff2a2b7a9ea2b2dd278fd9c0c0b827ac41bbc/src/hermes/commands/harvest/git.py#L287
is to get the current branch name/state. If the harvester would do sanity checks on the git output, `harvest_git` would be expected to fail here. This information is also passed to the context update
https://github.com/hermes-hmc/workflow/blob/f02ff2a2b7a9ea2b2dd278fd9c0c0b827ac41bbc/src/hermes/commands/harvest/git.py#L315-L323
.. but apparently does not end up in the context json, otherwise the test would fail ... what is going on here?

jkelling (changes requested):

test/hermes_test/commands/harvest/test_git.py
jkelling: The branch is passed to `HermesHarvestContext.updat_from` as a keyword arg and remian unused therein. I.e. the proposed mock works *until* we decide to record the branch in the harvest context.

sdruskat: > This information is also passed to the context update
> 
> https://github.com/hermes-hmc/workflow/blob/f02ff2a2b7a9ea2b2dd278fd9c0c0b827ac41bbc/src/hermes/commands/harvest/git.py#L315-L323
> 
> .. but apparently does not end up in the context json, otherwise the test would fail ... what is going on here?

Huh, must've slipped through review. The information *isn't* actually passed to the context. To pass it, either include it in the `dict` itself

```python
    ctx.update_from({ 
     '@context': [ 
         "https://doi.org/10.5063/schema/codemeta-2.0", 
         {'hermes': 'https://software-metadata.pub/ns/hermes/'} 
     ], 
  
     '@type': "SoftwareSourceCode", 
     'contributor': [contributor.to_codemeta() for contributor in git_contributors._all], 
     'branch': git_branch
 }) 
```

Or add it separately (which I'd prefer for reasons of readability) with `ctx.update('branch', git_branch)`.
This is a bug, tracked in

zyzzyxdonta (approved):
👍🏻

sdruskat:
Waiting for @jkelling's re-review before merging.

jkelling (approved):

PR #125, opened 2023-01-19
Implement metadata mapping for Invenio-based platforms

### LINKED ISSUE #58: Implement Invenio mapper
We need to implement a mapper from internal data model to whatever the pub repo accepts.

zyzzyxdonta:
Closes #58 

This pull request implements the "map" step of the deposit command for Invenio-based platforms. Currently, the code contains a lot of things that can be removed or simplified when more configuration options are available for hermes' processing step.

### DIFF
--- a/src/hermes/commands/workflow.py
+++ b/src/hermes/commands/workflow.py
@@ -124,7 +124,6 @@ def deposit():
         _log.error("You must run the process command before deposit")
         return 1
 
-    # TODO: How to write into the root of the context?
     codemeta_path = ContextPath("codemeta")
     with open(codemeta_file) as codemeta_fh:
         ctx.update(codemeta_path, json.load(codemeta_fh))
--- a/src/hermes/commands/workflow.py
+++ b/src/hermes/commands/workflow.py
@@ -147,7 +147,14 @@ def deposit():
         deposit_preparator = deposit_preparator_entrypoints[0].load()
         deposit_preparator(ctx)
 
-    # TODO: Metadata mapping
+    # Map metadata onto target schema
+    metadata_mapping_entrypoints = metadata.entry_points(
+        group="hermes.metadata_mapping",
+        name=deposition_platform
+    )
+    if metadata_mapping_entrypoints:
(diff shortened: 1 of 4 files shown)


src/hermes/commands/workflow.py
```diff
@@ -112,12 +112,22 @@ def deposit():
     Deposit processed (and curated) metadata
     """
     click.echo("Metadata deposition")
+    _log = logging.getLogger("cli.deposit")
```
led02, line 115: Does this work without the logger being configured? (I guess because the `cli` is defined... but not sure here)

led02 (approved):
Looks good for now. Personally, I'm a big fan of generative expressions. :) However, it might be a good idea to later extract certain mappings (e.g., lastName, givenName -> name) to a higher abstraction level. Personally, I'm a big fan of abstraction... ;)

src/hermes/commands/workflow.py
zyzzyxdonta: I don't think the name for the logger can be entirely made up. The docs say a hierarchical format is typical but they make no requirements. See [here](https://docs.python.org/3/library/logging.html#logging.getLogger).

led02: Yes, the name itself is in fact not constrained. However, in [src/hermes/config.py](config.py) I had to provide some basic configuration for the loggers. But I think the `cli` logger being defined there is sufficient for your code to actually produce log output. So should be fine here .:)

zyzzyxdonta: Oh I see. Yes, it does work. You can verify this by deleting you `.hermes` directory and then running `hermes deposit`. This should print "You must run the process command before deposit"

zyzzyxdonta: And this line is just a copy of these two:

https://github.com/hermes-hmc/workflow/blob/820c813b607d3ab0ad1ccb070c1611482cdadfd8/src/hermes/commands/workflow.py#L25

https://github.com/hermes-hmc/workflow/blob/820c813b607d3ab0ad1ccb070c1611482cdadfd8/src/hermes/commands/workflow.py#L57

Issue #117, opened 2023-01-18
Implement steps for Zenodo Sandbox that adds version to existing record


sdruskat:
This is the extended version of #61.
Ideally, we'll get this ready in time.

This could need, e.g.

- configurated concept DOI, or
- retrieving concept DOI based on project heuristics from Zenodo (i.e., an API "harvester")

zyzzyxdonta:
This should also include updating an existing record (i.e. only changing the metadata without creating a new version and changing files) as it also needs the configured DOI/url to publication/...

zyzzyxdonta:
Related: #14

PR #128, opened 2023-02-07
Deposit step for invenio-based platforms

### LINKED ISSUE #61: Implement deposit step for Zenodo Sandbox
- **Related:** #60 

We need to implement the deposit step to publish the metadata and optional software payload to the target pub repo.

It should be implemented in a way that makes it possible, easy even, to generalize its functionality to all InvenioRDM-based instances.

zyzzyxdonta:
This pull request implements the deposit step for Invenio-based platforms which will allow us to make first deposits on Zenodo Sandbox.

It can be tested by setting the environment variable `HERMES_INVENIO_AUTH_TOKEN` to your ["Personal Access Token"](https://sandbox.zenodo.org/account/settings/applications/) for Zenodo Sandbox and running `hermes deposit`, or passing `--auth-token YOUR-TOKEN` to `hermes deposit`. This will create a new publication on Zenodo Sandbox. The publication will contain only a `README.md` file containing some metadata. The README is created from scratch and is not the one from the repo that `hermes deposit` is run in.

**Note:** The access right of the publication is set to "open". This means that everyone with the correct link will see the publication. If you want to hide the content, set `"access_right": "closed"` before running the command.

https://github.com/hermes-hmc/workflow/blob/433a2b34c719a918d1c714453a8c87d92c5ff76f/src/hermes/commands/deposit/invenio.py#L141

Closes #61 

Creating a new version of an existing record will be implemented for #117.

### DIFF
--- a/src/hermes/cli.py
+++ b/src/hermes/cli.py
@@ -159,7 +159,3 @@ main.add_command(workflow.harvest)
 main.add_command(workflow.process)
 main.add_command(workflow.deposit)
 main.add_command(workflow.postprocess)
-
-
-if __name__ == "__main__":
-    main()
--- a/src/hermes/cli.py
+++ b/src/hermes/cli.py
@@ -159,3 +159,7 @@ main.add_command(workflow.harvest)
 main.add_command(workflow.process)
 main.add_command(workflow.deposit)
 main.add_command(workflow.postprocess)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -92,13 +92,6 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     deposit = response.json()
     _log.debug("Created deposit: %s", deposit["links"]["html"])
 
-    # ``deposit["metadata"]`` contains a ``prereserve_doi`` field which is a dict like
-    # this: ``{'doi': '10.5072/zenodo.1234567', 'recid': 1234567}``. The ``recid`` can
-    # be used to construct the URL to the record when published:
-    # ``f"https://sandbox.zenodo.org/record/{recid}"``.
-    # TODO: Use the URL and the DOI to update the data in the repo files before
-    # uploading them.
-
     # TODO: When updating existing releases, the files API can be used to delete
     # existing files. GET ``deposit["links"]["files"]`` for a list of ``files`` (names,
     # checksums, ...), then DELETE ``file["links"]["download"]`` for each ``file`` in
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -63,8 +63,9 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     invenio_path = ContextPath.parse("deposit.invenio")
     invenio_ctx = ctx[invenio_path]
 
-    # TODO: Get from environment or parameter to the deposit command. Click?
-    token = environ.get("HERMES_INVENIO_AUTH_TOKEN")
+    # TODO: Get from environment or parameter to the deposit command. This can be done
+    # using Click: https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables
+    token = environ["HERMES_INVENIO_AUTH_TOKEN"]
     click_ctx.session.headers["Authorization"] = f"Bearer {token}"
 
     # TODO: Get this from config or determine from some value (DOI, ...) in config.
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -42,6 +42,9 @@ def map_metadata(ctx: CodeMetaContext):
     metadata_path = ContextPath.parse("deposit.invenio.depositionMetadata")
     ctx.update(metadata_path, deposition_metadata)
 
+def deposit(ctx: CodeMetaContext):
+    pass
+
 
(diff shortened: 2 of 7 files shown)


src/hermes/cli.py
```diff
@@ -159,3 +159,7 @@ def main(ctx: click.Context, *args, **kwargs) -> None:
 main.add_command(workflow.process)
 main.add_command(workflow.deposit)
 main.add_command(workflow.postprocess)
+
+
+if __name__ == "__main__":
+    main()
```
led02, line 165: **question:** Why did you add another main block here? Is there another reason that just convenience? Since we have

- the script entry point to run `hermes`
- the `__main__.py` to run `python -m hermes`

I don't see the point here to add yet another main routine.

src/hermes/commands/deposit/invenio.py
```diff
@@ -4,17 +4,21 @@
...
+import logging
+from datetime import date, datetime
 
-import requests
+import click
 
 from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
 
+from .error import DepositionUnauthorizedError
```
led02, line 16: **nitpick:** I'm not a big fan of relative imports as they might end up doing weird stuff.
In addition, I'm not quite cerain whether we really want to import the class and not only the module - I'm a big fan of namespaces.
However, here this might be okay for reasons, but we should discuss whether some coding guidelines should be applied for similar cases.

```suggestion
from hermes.commands.deposit.error import DepositionUnauthorizedError
```

src/hermes/commands/deposit/invenio.py
```diff
@@ -43,13 +49,81 @@ def map_metadata(ctx: CodeMetaContext):
...
 
-def _request_json(url: str) -> dict:
-    """Request an URL and return the JSON response as dict."""
+def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
+    """Make a deposition on an Invenio-based platform.
+
+    This function can:
+
+    - Update the metadata of an existing record
+    - Update the metadata and files of an existing record y creating a new version
```
led02, line 58:
```suggestion
    - Update the metadata and files of an existing record by creating a new version
```

**concern:** But looking into the code, it says it can't.

src/hermes/commands/deposit/invenio.py
```diff
@@ -43,13 +49,81 @@ def map_metadata(ctx: CodeMetaContext):
...
+    # the list.
+
+    # Upload the files. We'll use the bucket API rather than the files API as it
+    # supports file sizes above 100MB.
+    bucket_url = deposit["links"]["bucket"]
+    file_name, file_content = _get_file_for_upload(deposition_metadata)
+
+    # TODO: When uploading real files, open them in binary mode and pass the handle to
+    # the data kwarg. For now we'll upload just this one file.
+    # TODO: Do we need to URL-encode the ``file_name`` or does this always work?
```
led02, line 111: **suggestion:** So many TODOs, maybe an issue tracker is a better place to keep track of them.

src/hermes/commands/workflow.py
```diff
@@ -107,13 +109,25 @@ def process():
...
 
+    # TODO: Better name than session?
+    # TODO: If this is needed in more places, it could be moved one level up.
+    click_ctx.session = requests.Session()
+    click_ctx.session.headers = {
+        "User-Agent": hermes_user_agent,
+    }
+    # TODO: This seems weird. Is there an option to put it directly into the context
+    # instead of as an argument?
+    click_ctx.auth_token = auth_token
```
led02, line 129: **note:** It should be available as `click_ctx.params["auth_token"]` (or similar, I guess).

src/hermes/utils.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: 2023 Helmholtz-Zentrum Dresden-Rossendorf (HZDR)
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# SPDX-FileContributor: David Pape
+
+from importlib.metadata import metadata
+
+
+hermes_metadata = metadata("hermes").json
```
led02, line 10: **question:** Why this extra conversion step here instead of using the PEP 314 directly?

led02 (commented):
Code looks good to me, but I have some open questions. Thanks for the great work.

Some parts of the code certainly need to be adapted, once I've been productive and the configuration infrastructure is in place...

src/hermes/cli.py
zyzzyxdonta: Oops, I completely missed the `__main__.py` 🤦🏻‍♂️ Reverted in 48c93e9adb39cf410ee16d9d975c8cdfe326b06c

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: Is cdee4a8bc362e42aa8746b9bd215f156b8dbcdd2 better?

src/hermes/utils.py
zyzzyxdonta: No reason. I changed it in 980270a27a6be1cef6bd86076281fc1d6e41e022

src/hermes/commands/workflow.py
zyzzyxdonta: Yes, thank you! fcd01ea9d58fcb168173a33009840a5e4212828e

led02 (approved):

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: I moved some stuff into issues.

PR #132, opened 2023-02-15
Feature/46 define and implement configuration strategy

led02:
Implement first configuration possibilities

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -12,3 +12,4 @@
 # Real Name <email@address> nickname
 # Real Name <email@address> Name, Real
 Jeffrey Kelling <j.kelling@hzdr.de> jkelling
+David Pape <d.pape@hzdr.de> David
--- a/hermes.toml
+++ b/hermes.toml
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2023 German Aerospace Center (DLR)
+#
+# SPDX-License-Identifier: CC0-1.0
+
 [harvest]
 from = [ "cff", "git" ]
 
--- a/hermes.toml
+++ b/hermes.toml
@@ -8,13 +8,9 @@ from = [ "cff", "git" ]
 [harvest.cff]
 validate = false
 
-[process]
-preprocess = [ "cff", "git" ]
-
 [deposit]
 mapping = "invenio"
 target = "invenio"
 
 [deposit.invenio]
 base_url = "https://sandbox.zenodo.org"
-auth_token = "UGOJ9fwRXUhshyWzQtqAqHKQRHlT6VPUrUNRV9v984zkjUvjhwFo8SnVYtXx"
--- a/hermes.toml
+++ b/hermes.toml
@@ -2,7 +2,7 @@
 from = [ "cff", "git" ]
 
 [harvest.cff]
-validate = true
+validate = false
 
 [process]
 preprocess = [ "cff", "git" ]
@@ -13,3 +13,4 @@ target = "invenio"
 
 [deposit.invenio]
 base_url = "https://sandbox.zenodo.org"
+auth_token = "UGOJ9fwRXUhshyWzQtqAqHKQRHlT6VPUrUNRV9v984zkjUvjhwFo8SnVYtXx"
--- a/src/hermes/cli.py
+++ b/src/hermes/cli.py
@@ -139,7 +139,10 @@ class WorkflowCommand(click.Group):
 
 
 @click.group(cls=WorkflowCommand, invoke_without_command=True)
-@click.option("--config", default=pathlib.Path('hermes.toml'), help="Configuration", type=pathlib.Path)
+@click.option(
+    "--config", default=pathlib.Path('hermes.toml'),
+    help="Configuration file in TOML format", type=pathlib.Path
+)
 @click.option("--deposit", is_flag=True, default=False)
 @click.option("--postprocess", is_flag=True, default=False)
 @click.option("--clean", is_flag=True, default=False)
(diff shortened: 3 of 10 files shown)


test/hermes_test/test_cli.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2022 German Aerospace Center (DLR)
 #
 # SPDX-License-Identifier: Apache-2.0
-
+import pathlib
```
zyzzyxdonta, line 4: Automatic import by IDE? :-) Should be below the SPDX-FileContributor

zyzzyxdonta (commented):
I haven't tried anything yet but here are some things I noticed.

test/hermes_test/test_cli.py
led02: Yes, indeed...

hermes.toml
```diff
@@ -0,0 +1,20 @@
...
+[process]
+preprocess = [ "cff", "git" ]
+
+[deposit]
+mapping = "invenio"
+target = "invenio"
+
+[deposit.invenio]
+base_url = "https://sandbox.zenodo.org"
+auth_token = "UGOJ9fwRXUhshyWzQtqAqHKQRHlT6VPUrUNRV9v984zkjUvjhwFo8SnVYtXx"
```
led02, line 20: This is of cause not a valid and working auth token (anymore)

src/hermes/config.py
```diff
@@ -4,78 +4,117 @@
...
-            config_toml = toml.load(config_file)
-            hermes_config = config_toml['tool'].get('hermes', {})
+        with open(config_path, 'r') as config_file:
+            hermes_config = toml.load(config_file)
             _config['hermes'] = hermes_config
             _config['logging'] = hermes_config.get('logging', _config['logging'])
 
-    except IOError:
-        pass
+    # Exceptions are currently handled gracefully. Is this okay?
```
zyzzyxdonta, line 78: I don't think this is a good idea. If the configuration file is not found, this means the user intended to configure something but messed up. In this case, hermes using its default behavior is probably the one thing they tried to avoid.

In the case where no file name is given and the default file is not found, I'm not sure what to do. Would it make sense to run hermes without any configuration? Probably for harvesting and such. For depositing data, it might not be possible (depending on how we set the default site for deposits). But the user also could have intended to create a file with the default name but accidentally misspelled it.

hermes.toml
zyzzyxdonta: I don't think it would be a good idea to allow users to configure the token in the config file. The config file will be commited to the repo and this will result in accidentally commited auth tokens. Only using command line options and environment variables should be the way to go.

src/hermes/cli.py
```diff
@@ -139,6 +139,7 @@ def _process_result(value: t.Any) -> t.Any:
 
 
 @click.group(cls=WorkflowCommand, invoke_without_command=True)
+@click.option("--config", default=pathlib.Path('hermes.toml'), help="Configuration", type=pathlib.Path)
```
zyzzyxdonta, line 142:
```suggestion
@click.option(
    "--config", default=pathlib.Path('hermes.toml'),
    help="Configuration file in TOML format", type=pathlib.Path
)
```

hermes.toml
led02: I totally agree. 😊

poikilotherm: I completely and strongly agree. These secrets should ideally be read from an env var or some file. All CI systems expose secrets as env vars, so probably that would be fine. It might be nice to configure the name of the env var to read, though.

zyzzyxdonta:
Another thing I thought about: Can we keep supporting pyproject.toml? This would be nice for people who have one anyways.

I.e. if a file is given via `--config` use it; else if `hermes.toml` exists, use this; else if `pyproject.toml` exists, look for `[tool.hermes]` in there.

led02:
> Another thing I thought about: Can we keep supporting pyproject.toml? This would be nice for people who have one anyways.

I think we opted against it for now. I indeed think it would be a great idea to support *different* project setup files, like `pyproject.toml`, `pom.xml`, `package.json` and what else. To be allowed to have this, I chose to store the loaded config in this awkward dictionary. For now, only supporting `pyproject.toml` would probably mislead the audience to think that Python projects are better supported. Hence, we decided to focus on the `hermes.toml` for now to stess that we are OS-agnostic.

src/hermes/commands/workflow.py
```diff
@@ -74,9 +75,19 @@ def process():
         _log.error("You must run the harvest command before process")
         return 1
 
-    # TODO: needs a lookup in future configuration to loop only over enabled harvesters
-    harvesters = metadata.entry_points(group='hermes.harvest')
-    for harvester in harvesters:
+    # Get all harvesters
+
+    # Get all harvesters
```
zyzzyxdonta, line 80:
```suggestion
    # Get all harvesters
```

.mailmap
```diff
@@ -12,3 +12,4 @@
 # Real Name <email@address> nickname
 # Real Name <email@address> Name, Real
 Jeffrey Kelling <j.kelling@hzdr.de> jkelling
+David Pape <d.pape@hzdr.de> David
```
zyzzyxdonta, line 15: 🙈 Sorry. Additionally, I changed my name now 😄

zyzzyxdonta:
> > Another thing I thought about: Can we keep supporting pyproject.toml? This would be nice for people who have one anyways.
> 
> I think we opted against it for now. I indeed think it would be a great idea to support _different_ project setup files, like `pyproject.toml`, `pom.xml`, `package.json` and what else.

Makes sense. I'll add an issue for this so we don't forget.

.mailmap
led02: No worries. I'll rather see this file as a record of contributions or similar... ;)

zyzzyxdonta (approved):
Great, I've got nothing more to add :-)

PR #144, opened 2023-03-16
Configure depositition through config file

zyzzyxdonta:
This pull requests removes a workaround that was introduced to make the deposit.invenio module configurable through the context object, and moves this functionality into the now existing hermes.toml config file.

### DIFF
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -15,11 +15,11 @@ import requests
 
 from hermes import config
 from hermes.commands.deposit.error import DepositionUnauthorizedError
+from hermes.error import MisconfigurationError
 from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
 from hermes.utils import hermes_user_agent
 
-_DEFAULT_SITE_URL = "https://sandbox.zenodo.org"
 _DEFAULT_DEPOSITIONS_API_PATH = "api/deposit/depositions"
 _DEFAULT_RECORD_SCHEMA_PATH = "api/schemas/records/record-v1.0.0.json"
 
@@ -37,7 +37,10 @@ def prepare_deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     invenio_path = ContextPath.parse("deposit.invenio")
     invenio_config = config.get("deposit").get("invenio", {})
 
-    site_url = invenio_config.get("site_url", _DEFAULT_SITE_URL)
+    site_url = invenio_config.get("site_url")
+    if site_url is None:
+        raise MisconfigurationError("invenio.site_url is not configured")
+
     record_schema_path = invenio_config.get("schema_paths", {}).get(
         "record", _DEFAULT_RECORD_SCHEMA_PATH
     )
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -99,7 +99,10 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
 
     existing_record_url = None
 
-    site_url = invenio_config.get("site_url", _DEFAULT_SITE_URL)
+    site_url = invenio_config.get("site_url")
+    if site_url is None:
+        raise MisconfigurationError("deposit.invenio.site_url is not configured")
+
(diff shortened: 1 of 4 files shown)


src/hermes/commands/deposit/invenio.py
```diff
@@ -12,10 +12,15 @@
 
 import click
 
+from hermes import config
 from hermes.commands.deposit.error import DepositionUnauthorizedError
 from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
 
+_DEFAULT_SITE_URL = "https://sandbox.zenodo.org"
```
led02, line 20: Is it really the "sandbox" that we want to default to?

src/hermes/commands/deposit/invenio.py
```diff
@@ -69,16 +78,24 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
+    invenio_config = config.get("deposit").get("invenio", {})
     invenio_path = ContextPath.parse("deposit.invenio")
     invenio_ctx = ctx[invenio_path]
 
     if not click_ctx.params["auth_token"]:
         raise DepositionUnauthorizedError("No auth token given for deposition platform")
-    click_ctx.session.headers["Authorization"] = f"Bearer {click_ctx.params['auth_token']}"
+
+    http_client = click_ctx.obj.get("http_client")
+    http_client.headers["Authorization"] = f"Bearer {click_ctx.params['auth_token']}"
```
led02, line 89: Clever idea hiding the complexity of requests here. However, this might have unexprected side effects (e.g., if you want multiple deposits or other steps use the http_client, too).

led02 (commented):
**note:** There are some questions that we should discuss before I can reasonably approve this PR. Sorry!

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: Good point. In the worst case, we could leak the authentication token to some other site which could then, based on the hermes user agent, guess where to apply this token (some site targeted by hermes). I'll come up with something different.

zyzzyxdonta: I made the requests.Session local in 5236f9a84ae2d791c901ff1ae1554142e6236198

src/hermes/commands/deposit/invenio.py
led02: Yes, that sounds good. However, as long as we're not released, it's okay to faciliate testing... or we need to also start having example configs / test configs.

To sum up: This change would not block an approval from my side ;)

zyzzyxdonta: I'm throwing an exception now in 2bd83ad35029840e06950d388c9628c577d5754c

src/hermes/commands/deposit/invenio.py
```diff
@@ -28,17 +33,21 @@ def prepare_deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
+    invenio_config = config.get("deposit").get("invenio", {})
 
-    invenio_ctx = ctx[invenio_path]
-    # TODO: Get these values from config with reasonable defaults.
-    recordSchemaUrl = f"{invenio_ctx['siteUrl']}/{invenio_ctx['schemaPaths']['record']}"
+    site_url = invenio_config.get("site_url", _DEFAULT_SITE_URL)
+    record_schema_path = invenio_config.get("schema_paths", {}).get(
+        "record", _DEFAULT_RECORD_SCHEMA_PATH
+    )
+    record_schema_url = f"{site_url}/{record_schema_path}"
```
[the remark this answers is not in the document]
led02: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

Again: It's not something that would not block my approval. :)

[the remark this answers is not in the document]
zyzzyxdonta: I didn't know this existed and, to be honest, it looks a bit error-prone 😄 

```python
>>> urljoin("https://example.com/test", "foo.json")
'https://example.com/foo.json'
*>>> urljoin("https://example.com/test/", "foo.json")
'https://example.com/test/foo.json'
```

I'll keep it as is 😁

PR #146, opened 2023-03-20
Resolve internal Invenio Record ID of latest record

### LINKED ISSUE #142: Resolve InvenioRDM internal latest version ID
Blocks for #117, #130.

The following strategy is used:

- If an invenio ID is configured, use this to retrieve the latest verison.
- If a (concept) DOI is configured, use this to retrieve the latest version.
- If a codemeta `identifier` is present in the metadata, assume it to be a DOI and use this to retrieve latest version.
- If none of this exists, reserve a new deposition.

led02:
Closes #142...


src/hermes/commands/deposit/invenio.py
```diff
@@ -132,6 +137,90 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
+    Anyway, the record id will always be use to resolve the latest version.
+
+    If some of the resolution steps fail or produce an unexpected result, a ValueError will be thrown.
+
+    :param ctx: The context for which the record id should be resolved.
+    :return: The Invenio record id.
+    """
+
+    cfg = config.get('deposit').get('invenio', {})
+    base_url = cfg.get('base_url', 'https://sandbox.zenodo.org')
```
sdruskat, line 157: Wouldn't it be better to handle (or keep as `None`) non-existent `base_url`s rather than configuring a fallback to Zenodo Sandbox?

src/hermes/commands/deposit/invenio.py
```diff
@@ -132,6 +137,90 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
+    record_id = cfg.get('record_id')
+
+    # Check if we configured an Invenio record ID (of the concept...)
+    if record_id is None:
+        doi = cfg.get('doi')
+        if doi is None:
+            # TODO: There might be more seantic in the codemeta.identifier... (also see schema.org)
+            identifier = ctx['codemeta.identifier']
+            if identifier.startswith('https://doi.org/'):
+                doi = identifier[16:]
```
sdruskat, line 167:
```suggestion
            if identifier.startswith('https://doi.org/'):
                doi = identifier[16:]
            elif identifier.startswith('http://dx.doi.org/'):
                doi = identifier[18:]
```

This catches the other common (older) resolver URL which is still found in abundance. As both may come as HTTP *and* HTTPS URLs, I've created #147 to track this.

As CodeMeta `identifier` can also contain the actual DOI (excluding the resolver URL), these cases should be handled as well. Created #148 for this.

sdruskat (approved):
Looks good to me, thanks. 🍰 

Added some comments and one suggestion, also created #147 and #148 during review.
Feel free to merge as is, or merge the suggestion first.

sdruskat:
⚠️ Wait for @zyzzyxdonta's review before merging.

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: Also, I changed `base_url` to `site_url` in an earlier PR because it made more sense to me* and it was merged that way. So here, it should be `site_url` too.

*) This is the site the user goes to and the user doesn't care that it is the base for the API URLs and so on.

zyzzyxdonta: For a missing `site_url`, I introduced `MisconfigurationException` in #144. Maybe that would be an option here.

src/hermes/commands/deposit/invenio.py
```diff
@@ -132,6 +137,90 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     _log.info("Published record: %s", record["links"]["record_html"])
 
 
+def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> str:
```
zyzzyxdonta, line 140: This function returns `None` if the last check fails

led02: That's correct and intended... I was not quite sure how to eventually handle the error here. It's also related to your comment below

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: Now merged and available on develop

src/hermes/commands/deposit/invenio.py
led02: Now, a ValueError is raised... needs to be cleaned up further, maybe raise a custom error or so?

PR #149, opened 2023-03-23
Write valid CodeMeta

### LINKED ISSUE #126: Collect contexts separately and write them together only when writing out codemeta.json
Once #122 is merged, we're writing out invalid CodeMeta JSON, as the hermes `@context` isn't patched through to the CodeMeta file.
This is because the processor writes only contributors and branch to the CodeMetaContext, and not the `hermes` `@context`, which is being added in the git harvester. Witness:

> https://github.com/hermes-hmc/workflow/blob/820c813b607d3ab0ad1ccb070c1611482cdadfd8/src/hermes/commands/harvest/git.py#L316-L319

and

> https://github.com/hermes-hmc/workflow/blob/f1ff56cc62b49651ba7b81a053244ac5395b9e28/src/hermes/commands/process/git.py#L72-L75

In theory, we wouldn't need a git processor at all at this stage, as the harvester currently does 3 things: 

1. Collect cnotributors
2. Collect branch
3. Add the JSON-LD contexts (i.e., re-add the CodeMeta `@context` as default context, this is already being done in the CFF harvester through conversion; add the named `hermes` `@context`).

As is, if we take out the processor, merging in below crashes, as we have one string `@context` (CodeMeta) and try to add a dict item (named hermes `@context`).

> https://github.com/hermes-hmc/workflow/blob/820c813b607d3ab0ad1ccb070c1611482cdadfd8/src/hermes/commands/workflow.py#L89

(Note also the weirdly inverted logic, as this line looks as if we're merging something into the harvesting context, while the `merge_from` method writes data from `other` into the context data passed as first argument.)

The same would happen if we added the new context (hermes) in the processor.

### Proposition

- Collect newly added contexts in *harvesters* (we want valid contexts here already).
- Collect into a new collection (similar to `tags`).
- Pass one this collection (`contexts`) to other `HermesContexts`.
- Once processing is done in the workflow (`workflow.py`!), handle the `@context`s:
    - E.g., put CodeMeta as default `@context`, deduplicate other contexts by value (URI) and update keys
        - Our own `hermes` context must be defined and made available, and documented as well
    - Write out contexts

sdruskat:
This PR fixes #126.

- Provide API for collecting additional linked data contexts during harvesting
- Merge linked data contexts into CodeMetaContext during processing
- Prepare valid CodeMeta by dumping uncontrolled linked data contexts (e.g., from conversion of CFF to CodeMeta) from data, adding merged contexts, and setting the `@type` property explicitly.

### DIFF
--- a/src/hermes/commands/harvest/git.py
+++ b/src/hermes/commands/harvest/git.py
@@ -313,14 +313,10 @@ def harvest_git(click_ctx: click.Context, ctx: HermesHarvestContext):
     _audit_contributors(git_contributors, logging.getLogger('audit.git'))
 
     ctx.update_from({
-        '@context': [
-            "https://doi.org/10.5063/schema/codemeta-2.0",
-            {'hermes': 'https://software-metadata.pub/ns/hermes/'}
-        ],
-
         '@type': "SoftwareSourceCode",
         'contributor': [contributor.to_codemeta() for contributor in git_contributors._all],
     }, git_branch=git_branch)
+    ctx.add_context(('hermes', 'https://software-metadata.pub/ns/hermes/'))
     ctx.update('hermes:gitBranch', git_branch)
 
     try:
--- a/src/hermes/commands/harvest/git.py
+++ b/src/hermes/commands/harvest/git.py
@@ -315,7 +315,6 @@ def harvest_git(click_ctx: click.Context, ctx: HermesHarvestContext):
     ctx.update('contributor',
                [contributor.to_codemeta() for contributor in git_contributors._all],
                git_branch=git_branch)
-    ctx.add_context(('hermes', 'https://software-metadata.pub/ns/hermes/'))
     ctx.update('hermes:gitBranch', git_branch)
 
     try:
--- a/src/hermes/commands/harvest/git.py
+++ b/src/hermes/commands/harvest/git.py
@@ -312,10 +312,9 @@ def harvest_git(click_ctx: click.Context, ctx: HermesHarvestContext):
 
     _audit_contributors(git_contributors, logging.getLogger('audit.git'))
 
-    ctx.update_from({
-        '@type': "SoftwareSourceCode",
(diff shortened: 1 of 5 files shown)


sdruskat:
Ready for review @led02, @zyzzyxdonta, this time for real.

src/hermes/commands/harvest/git.py
```diff
@@ -312,15 +312,10 @@ def harvest_git(click_ctx: click.Context, ctx: HermesHarvestContext):
...
-            {'hermes': 'https://software-metadata.pub/ns/hermes/'}
-        ],
-
-        '@type': "SoftwareSourceCode",
-        'contributor': [contributor.to_codemeta() for contributor in git_contributors._all],
-    }, git_branch=git_branch)
+    ctx.update('contributor',
+               [contributor.to_codemeta() for contributor in git_contributors._all],
+               git_branch=git_branch)
+    ctx.add_context(('hermes', 'https://software-metadata.pub/ns/hermes/'))
```
zyzzyxdonta, line 318: Adding the hermes context isn't really git harvester-specific. I think this should happen elsewhere. Maybe the hermes context should exist by default in all HermesContext objects.

sdruskat: In this specific case, yes. However, I think all harvesters should be able to add arbitrary contexts.

zyzzyxdonta: Absolutely 👍🏻

sdruskat: Will move adding the hermes context to the (tada) `HermesContext` and chuck the URL into a constant.

zyzzyxdonta: Cool 👍🏻

I don't think removing the context is necessary, either. (If it was, every context that was added would need to be checked, not only the hermes context.) Maybe at some point in the future we could have a custom write function that does this before writing the data to disk. But before doing this we should look into some json-ld libraries...

zyzzyxdonta (approved):

PR #151, opened 2023-03-29
Use the correct software license when depositing files to Invenio instance

### LINKED ISSUE #150: Use correct software license in deposit metadata mapping
Currently, Apache-2.0 is always used. We need to get this information from the CFF harvester (or from elsewhere) and map it correctly.

zyzzyxdonta:
This pull request fixes the metadata mapping for Invenio-based depositions so that the correct license is used.

The implementation assumes that the license field in the codemeta is a URL whose path ends in a license identifier that can be found on the Invenio instance.

Closes #150

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -15,6 +15,7 @@ message: >-
   If you use this software, please cite it using the
   metadata from this file.
 version: "proof-of-concept"
+license: "Apache-2.0"
 abstract: "Proof-of-concept implementation of the HERMES workflow."
 type: software
 authors:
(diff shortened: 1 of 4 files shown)


led02 (approved):
*hint:* Might improve from a little more documentation, but looks fine for me now.

We still need to discuss, what Exceptions when to throw and where to catch. But this is nothing for the PR, I think...

sdruskat (approved):
Thanks @zyzzyxdonta, looks good to me, only 1 typo to fix. Feel free to merge yourself please.

PR #152, opened 2023-03-30
Implement new version deposition on Zenodo Sandbox

### LINKED ISSUE #117: Implement steps for Zenodo Sandbox that adds version to existing record

led02:
Closes #117.

Demo: https://sandbox.zenodo.org/record/1177872

### DIFF
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -38,7 +38,8 @@ def prepare_deposit(click_ctx: click.Context, ctx: CodeMetaContext):
 
     invenio_path = ContextPath.parse("deposit.invenio")
     invenio_config = config.get("deposit").get("invenio", {})
-    _ = _resolve_latest_invenio_id(ctx)
+    rec_id, rec_meta = _resolve_latest_invenio_id(ctx)
+    ctx.update(invenio_path['latestRecord'], {'id': rec_id, 'metadata': rec_meta})
 
     site_url = invenio_config.get("site_url")
     if site_url is None:
@@ -119,6 +120,10 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
         )
 
     deposition_metadata = invenio_ctx["depositionMetadata"]
+    latest_metadata = invenio_ctx["latestRecord"]["metadata"]
+    if deposition_metadata.get("version") == latest_metadata.get("version"):
+        raise ValueError("Deposited version already version to deposit.")
+
     response = session.post(
         deposit_url,
         json={"metadata": deposition_metadata}
@@ -164,7 +169,7 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     _log.info("Published record: %s", record["links"]["record_html"])
 
 
-def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> str:
+def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> (str, dict):
     """
     Using the given configuration and metadata, figure out the latest record id.
 
@@ -177,7 +182,7 @@ def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> str:
     If any of the resolution steps fail or produce an unexpected result, a ValueError will be thrown.
 
     :param ctx: The context for which the record id should be resolved.
-    :return: The Invenio record id.
+    :return: The Invenio record id and the metadata of the record
     """
 
     invenio_config = config.get('deposit').get('invenio', {})
@@ -205,7 +210,12 @@ def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> str:
         # If we got a record id by now, resolve it using the Invenio API to the latests record.
         return _invenio_resolve_record_id(site_url, record_id)
 
-    raise ValueError("Could not figure out how to retrieve record id.")
+    # TODO figure out whether an initial deposit is what we want...
+    trigger_initial_deposit = False
+    if trigger_initial_deposit:
+        return None, {}
+    else:
+        raise ValueError("Could not figure out how to retrieve record id.")
 
 
 def _invenio_resolve_doi(site_url, doi) -> str:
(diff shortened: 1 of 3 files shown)


src/hermes/commands/deposit/invenio.py
```diff
@@ -111,18 +110,32 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
-    if existing_record_url is not None:
-        # TODO: Get by calling new version on latest existing record
-        deposit_url = None
-        raise NotImplementedError(
-            "At the moment, hermes can not create new versions of existing records"
+    deposition_metadata = invenio_ctx["depositionMetadata"]
+    try:
+        latest_metadata = invenio_ctx["latestRecord"]["metadata"]
+        if deposition_metadata.get("version") == latest_metadata.get("version"):
+            raise ValueError("Deposited version already version to deposit.")
```
zyzzyxdonta, line 117:
```suggestion
            raise ValueError("Version already deposited.")
```

Something like this

src/hermes/commands/deposit/invenio.py
```diff
@@ -205,7 +218,12 @@ def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> str:
         # If we got a record id by now, resolve it using the Invenio API to the latests record.
         return _invenio_resolve_record_id(site_url, record_id)
 
-    raise ValueError("Could not figure out how to retrieve record id.")
+    # TODO figure out whether an initial deposit is what we want...
+    trigger_initial_deposit = False
+    if trigger_initial_deposit:
+        return None, {}
+    else:
+        raise ValueError("Could not figure out how to retrieve record id.")
```
zyzzyxdonta, line 226: Whether we want to trigger an initial deposit should not influence the result of this function. I would always return the same thing (`None, {}` is fine, I think) and move this decision elsewhere.

led02: Yeah, totally agree. Even more as this is obviously bloat code

led02:
> `hermes deposit` currently fails if no `record_id` is configured. Is this the right thing to do? I don't want tell hermes to make a new deposition by setting `record_id` to a magic number 😁

Well... this is because you did not agree to do an initial deposit (change line 222 in [src/hermes/commands/deposit/invenio.py](https://github.com/hermes-hmc/workflow/pull/152/files/3ee43bccc8da1f269300a3ef1187cbe8435882da#diff-63a871d6c34c1ad1f7789cbc110534eaf6518dfd3402a9a4314c8f3f35e376b2)).
This behaviour is also due to sandbox usage as the DOI resolver does not work for this situation, otherwise a given DOI should be resolved to a record_id and the new version can be deposited.

src/hermes/commands/deposit/invenio.py
```diff
@@ -111,18 +110,30 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
+        if deposition_metadata.get("version") == latest_metadata.get("version"):
+            raise ValueError("Version already deposited.")
+
+        record_id = invenio_ctx["latestRecord"]["id"]
+    except KeyError:
+        record_id = None
+
+    if record_id is not None:
+        deposit_url += f'/{record_id}/actions/newversion'
+        response = session.post(deposit_url)
```
zyzzyxdonta, line 125: This response needs checking. If the request didn't succeed it was very likely because we are not the owner of this existing record. (I noticed this when trying to use your `record_id` and my `auth_token`.)

Currently we then fail in line 128 with `KeyError: 'links'` on the `old_deposit`.

src/hermes/commands/deposit/invenio.py
```diff
@@ -368,7 +383,7 @@ def _codemeta_to_invenio_deposition(metadata: dict) -> dict:
         "grants": None,
         "subjects": None,
         # TODO: Get this from config
```
zyzzyxdonta, line 385:
```suggestion
```

src/hermes/commands/deposit/invenio.py
```diff
@@ -158,13 +169,14 @@ def deposit(click_ctx: click.Context, ctx: CodeMetaContext):
...
         _log.error(f"Could not publish deposit via {publish_url!r}")
+        _log.debug(response.text)
         click_ctx.exit(1)
 
     record = response.json()
     _log.info("Published record: %s", record["links"]["record_html"])
 
 
-def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> str:
+def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> (str, dict):
```
zyzzyxdonta, line 179:
```suggestion
def _resolve_latest_invenio_id(ctx: CodeMetaContext) -> Tuple[str, dict]:
```

From Pylance:

>Tuple expression not allowed in type annotation
>  Use Tuple[T1, ..., Tn] to indicate a tuple type or Union[T1, T2] to indicate a union type

(needs `from typing import Tuple`)

src/hermes/commands/deposit/invenio.py
```diff
@@ -233,7 +248,7 @@ def _invenio_resolve_doi(site_url, doi) -> str:
     return record_id
 
 
-def _invenio_resolve_record_id(site_url: str, record_id: str) -> str:
+def _invenio_resolve_record_id(site_url: str, record_id: str) -> (str, dict):
```
zyzzyxdonta, line 251:
```suggestion
def _invenio_resolve_record_id(site_url: str, record_id: str) -> Tuple[str, dict]:
```

src/hermes/commands/deposit/invenio.py
led02: Well, I hope this is enough of a fix for now...

zyzzyxdonta: Looks good to me now 👍🏻

zyzzyxdonta (approved):
Works 👍🏻

PR #160, opened 2023-04-24
Allow users to configure invenio communities

### LINKED ISSUE #157: Configure Invenio communities for publication
Invenio-based platforms allow a kind of curation via "Communities" in which publications can be listed. Currently, hermes does not have the ability to add communities to a publication. This should be made possible via a configuration option for the deposition plugin.

zyzzyxdonta:
Closes #157

### DIFF
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -23,6 +23,7 @@ from hermes.model.path import ContextPath
 from hermes.utils import hermes_user_agent
 
 _DEFAULT_LICENSES_API_PATH = "api/licenses"
+_DEFAULT_COMMUNITIES_API_PATH = "api/communities"
 _DEFAULT_DEPOSITIONS_API_PATH = "api/deposit/depositions"
 _DEFAULT_RECORD_SCHEMA_PATH = "api/schemas/records/record-v1.0.0.json"
 
@@ -33,8 +34,8 @@ def prepare_deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     """Prepare the Invenio deposit.
 
     In this case, "prepare" means download the record schema that is required by
-    Invenio instances and checking whether we have a valid license identifier that is
-    supported by the instance.
+    Invenio instances and checking whether we have a valid license identifier and
+    communities (if any are configured) that are supported by the instance.
     """
 
     invenio_path = ContextPath.parse("deposit.invenio")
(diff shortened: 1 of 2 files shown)


src/hermes/commands/deposit/invenio.py
```diff
@@ -68,6 +69,13 @@ def prepare_deposit(click_ctx: click.Context, ctx: CodeMetaContext):
     license = _get_license_identifier(ctx, licenses_api_url)
     ctx.update(invenio_path["license"], license)
 
+    communities_api_path = invenio_config.get("api_paths", {}).get(
+        "communities", _DEFAULT_COMMUNITIES_API_PATH
+    )
+    communities_api_url = f"{site_url}/{communities_api_path}"
+    communities = _get_community_identifiers(ctx, communities_api_url)
+    ctx.update(invenio_path["communities"], communities)
```
led02, line 77: **nitpick:** I smell some duplicate code here... Probably fine for now, might be subject to further abstraction later on.

src/hermes/commands/deposit/invenio.py
```diff
@@ -438,3 +446,34 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
...
+    identifiers can not be found on the site, a :class:`MisconfigurationError` is
+    raised.
+    """
+
+    communities = config.get("deposit").get("invenio", {}).get("communities")
+    if communities is None:
+        return None
+
+    session = requests.Session()
+    session.headers = {"User-Agent": hermes_user_agent}
```
led02, line 465: **question:** I like the idea of a custom user agent... maybe we should move the network handling code to a common submodule (however, out-of-scope for poc).

led02 (approved):
Some potential improvement for later versions exists, but this is fine for now.

However, we should consider some additional refactorings for *beta*:
- Network helpers API
- Better configuration file handling (e.g., providing defaults etc.)
- Common API for checking / matching against a ReST endpoint

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: Yes, I noticed this, too. Even in the `_get_community/license_identifier(s)` there are some duplications

src/hermes/commands/deposit/invenio.py
zyzzyxdonta: Agreed. This is the second time this has come up so I'll write an issue for this.

zyzzyxdonta: https://github.com/hermes-hmc/workflow/issues/165

PR #171, opened 2023-05-03
Make access_rights for Invenio deposit configurable

zyzzyxdonta:


src/hermes/commands/deposit/invenio.py
```diff
@@ -484,3 +488,70 @@ def _get_community_identifiers(ctx: CodeMetaContext, communities_api_url: str):
...
+
+    if license is None and access_right in ["open", "embargoed"]:
+        raise MisconfigurationError(
+            f"With access_right {access_right}, a license is required."
+        )
+
+    if access_right == "closed":
+        pass
+
+    return access_right, embargo_date, access_conditions
```
led02, line 557: *note:* I see lots of potential for abstraction here... we have a number of dependent fields, each is checked for being required and if there is a valid value given.
*conclusion:* We should do a proper code review after PoC and collect our findings

led02 (approved):

PR #173, opened 2023-05-03
Add '--initial' flag for deposition step

led02:
Otherwise, an initial deposition is not done.

### DIFF
--- a/src/hermes/commands/workflow.py
+++ b/src/hermes/commands/workflow.py
@@ -148,6 +148,7 @@ def curate(click_ctx: click.Context):
 
 
 @click.group(invoke_without_command=True)
+@click.option("--initial", is_flag=True, default=False)
 @click.option(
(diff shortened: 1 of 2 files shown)


src/hermes/commands/workflow.py
```diff
@@ -148,6 +148,10 @@ def curate(click_ctx: click.Context):
 
 
 @click.group(invoke_without_command=True)
+@click.option(
+    "--initial", is_flag=True, default=False,
+    help="Make initial deposition. This is required if you do not have configured any reference to a previous version."
```
jkelling, line 153: Maybe do not say "Make inital deposition", because this is not implemented behavior. Rather:
```
Allow initial deposition if no previous version exists in target repository. Otherwise only an existing, configured upstream record may be updated.
```

jkelling (approved):

Issue #138, opened 2023-02-21
Make logging nicer and where to store its output


poikilotherm:
Some thoughts from hacking on the tutorial #110:

- It seems somewhat obscure to me why we drop `hermes.log` and `hermes-audit.md` not inside `.hermes` or `.hermes/logs` but within the project's root. Wouldn't it make sense to leave these debug/trace files in our "build directory"?
- The log on the console is Markdown. For me as a devop this looks very weird. Can we make this cleaner and oriented to how other CLI tools do logging to console?
- I hacked on the deRSE23 branch to include handling for the `HTTPError` exceptions thrown when deposits fail to print the actual response and not show a user stacktraces. As far as I can tell, we don't have mechanics in place to handle any bubbling exceptions, which might be a good idea to change.

FObersteiner:
Bump. 

Two things that would appear to be useful improvements to me:
- use an ISO8601 timestamp when logging, instead of Unix time (which is not so human-friendly)
- in CI, more verbose output to the console would be nice - I'd rather not jump extra hoops to access the hermes.log from the CI runs

PR #180, opened 2023-05-17
Boil down first CI tutorial

sdruskat:
Tihs PR prepares a working tutorial for using GitHub Zenodo Sandbox publication via a template.

Issue #179, opened 2023-05-17
Clean up CLI UI


sdruskat:
Consider using vanilla Click subcommands instead of groups and the custom `WorkflowCommand` mechanism, and additionally add `chain=True` to enable Maven-like subcommand chaining:

`hermes clean harvest process curate`

The only thing we'd need to enable beyond the current implementation is "curation skipping" for cases where users don't want an extra curation step. Could be done via `hermes deposit --silent` or similar.

To test: Does chaining with arguments work correctly, e.g., `hermes clean harvest process curate deposit --auth-token=foo --file=README.md postprocess`?

poikilotherm:
This is related and/or a duplicate of #137

sdruskat:
Closing as duplicate of #137.

Issue #137, opened 2023-02-21
Straighten out CLI usage


poikilotherm:
This is something I noticed while working on #110 

A) It is pretty convenient to do a two step deposit by first running "hermes" and then "hermes deposit". Maybe it is not necessary to have "hermes --deposit", "hermes --postprocess" etc. This might also help the code structure, as it's kind of hard to integrate Click options and arguments with these flags, as it requires manual parsing instead of relying on Click decorator mechanics. I did also encounter some errors about unknown options etc.

B) I noticed that the help is showing the option to have another command after the first:
```
Usage: hermes deposit [OPTIONS] COMMAND [ARGS]...
```
I'm not sure this is on purpose to concat commands or a bug because of our special handling of commands.

I definitely ran into a problem with this. When trying to get file artifacts going as arguments, running `hermes --deposit --auth-token XXX --file YYY`, resulted in a "command not known" error.

PR #181, opened 2023-05-17
Improve log file locations

poikilotherm:
Relates to #138

1. Make sure the parent directories of the configured log file locations exist (so we can use them) - or else create them
2. Move default location of hermes log into cache dir

### DIFF
--- a/src/hermes/cli.py
+++ b/src/hermes/cli.py
@@ -85,7 +85,7 @@ class WorkflowCommand(click.Group):
         """
 
         # Get the user provided working dir from the --path option or default to current working directory.
-        working_path = ctx.params.get('path', pathlib.Path.cwd()).absolute()
+        working_path = ctx.params.get('path').absolute()
 
         configure(ctx.params.get('config').absolute(), working_path)
         init_logging()
--- a/src/hermes/cli.py
+++ b/src/hermes/cli.py
@@ -84,13 +84,16 @@ class WorkflowCommand(click.Group):
         :param ctx: Context for the command.
         """
 
-        configure(ctx.params.get('config').absolute())
+        # Get the user provided working dir from the --path option or default to current working directory.
+        working_path = ctx.params.get('path', pathlib.Path.cwd()).absolute()
+
+        configure(ctx.params.get('config').absolute(), working_path)
         init_logging()
         log_header(None)
 
         audit_log = logging.getLogger('audit')
         audit_log.info("# Running Hermes")
-        audit_log.info("Running Hermes command line in: %s", ctx.params.get('path', pathlib.Path.cwd()).absolute())
+        audit_log.info("Running Hermes command line in: %s", working_path)
         audit_log.debug("")
         audit_log.debug("Invoked `%s` with", ctx.invoked_subcommand or self.name)
         audit_log.debug("")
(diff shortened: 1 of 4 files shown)


jkelling (approved):

jkelling (changes requested):
Let us discuss the naming of the `--path` option. (have to post this comment to remove my approving review.)

src/hermes/cli.py
```diff
@@ -84,13 +84,16 @@ def invoke(self, ctx: click.Context) -> t.Any:
         :param ctx: Context for the command.
         """
 
-        configure(ctx.params.get('config').absolute())
+        # Get the user provided working dir from the --path option or default to current working directory.
+        working_path = ctx.params.get('path', pathlib.Path.cwd()).absolute()
```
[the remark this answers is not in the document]
poikilotherm: "-C" sounds kinda weird to me, but whatever. Could you leave a comment at #137 and/or #179 about this? Seems out of scope for this PR.

poikilotherm:
> Let us discuss the naming of the `--path` option. (have to post this comment to remove my approving review.)

I agree we should talk about this. Can we please keep this out of this PR? This one is relevant for properly setting up folks with #110 / #180 for the DataCite meeting. IMHO changing the path command line option seems out of scope for this PR which is *only* about adapting the logging paths.

jkelling (approved):
The option-name issue is indeed outside of the scope, as the option existed before, I just noticed it here first.

src/hermes/cli.py
jkelling, line 88: The path option has a default, which should be used here (`./`, i.e. the same)
```suggestion
        working_path = ctx.params.get('path').absolute()
```

jkelling (changes requested):

src/hermes/config.py
```diff
@@ -121,6 +129,10 @@ def init_logging():
     if _loggers:
         return
 
+    # Make sure the directories to hold the log files exists (or else create)
+    _config['logging']['handlers']['logfile']['filename'].parent.mkdir(exist_ok=True, parents=True)
+    _config['logging']['handlers']['auditfile']['filename'].parent.mkdir(exist_ok=True, parents=True)
+
```
led02, line 135: These will fail if there is something different than the default value configured (as this usually will be a string not a path...)

```suggestion
    # Make sure the directories to hold the log files exists (or else create)
    pathlib.Path(_config['logging']['handlers']['logfile']['filename']).parent.mkdir(exist_ok=True, parents=True)
    pathlib.Path(_config['logging']['handlers']['auditfile']['filename']).parent.mkdir(exist_ok=True, parents=True)

```

src/hermes/config.py
```diff
@@ -68,6 +70,12 @@ def configure(config_path: pathlib.Path):
     if 'hermes' in _config and _config['hermes']:
         return
 
+    # Load sane default paths for log files before potentially overwritting via configuration
+    _config['logging']['handlers']['logfile']['filename'] = \
+        working_path / HermesContext.hermes_cache_name / "hermes.log"
+    _config['logging']['handlers']['auditfile']['filename'] = \
+        working_path / HermesContext.hermes_cache_name / "audit.log"
+
```
led02, line 78: **nag:** I don't like this very much, however I have not better idea at all how to improve this right now.

Might be subject to a discussion for alpha: How should our configuration system work (especially in combination with plugins...)

led02 (changes requested):
Mostly okay, see not for requested change.

src/hermes/config.py
poikilotherm: Maybe. As you say, let's discuss this later.

PR #190, opened 2023-06-23
Replace `convert-codemeta` dependency with forked code

poikilotherm:
Avoid adding very large and unnecessary dependencies like `numpy` and `pandas` to an installation of `hermes` (which takes time, eats cache space and bloats Docker images)


jkelling:
Does this close by #190 ? It should be linked either way.

pyproject.toml
```diff
@@ -69,7 +69,7 @@ sphinxext-opengraph = "^0.6.3"
 sphinx-click = "^4.4.0"
 sphinxcontrib-mermaid="^0.8.1"
 sphinx-togglebutton="^0.3.2"
-reuse = "^1.0.0"
+reuse = "^1.1.2"
```
jkelling, line 72: Is this update required to make this patch work?

I think we decided to keep changes in `pyproject.toml` and  `poetry.lock` to a minimum. Do we allow package updates by PR side-effect?

src/hermes/commands/harvest/util/validate_codemeta.py
zyzzyxdonta, line 1: Do we want to stay compatible with the original library so that we can switch back later? If not, then I would remove the `print` calls and maybe return a message at the end or raise an exception.

Is this harvest-specific? I feel like it could be used in different places, so I would move it into a more top-level module like `hermes.utils` (or maybe even `hermes.validation`)

pyproject.toml
led02: **opinion:** Well, I think as there is a *proper* change here, it's okay to also update the dependencies...

led02 (approved):

PR #217, opened 2023-09-29
Add --no-show-signature flag to git

### LINKED ISSUE #216: Disable git log signature verification
Hermes' git harvester calls the `git log` command to extract metadata from the git history of a repository. Depending on the git configuration, this might cause git to try and verify the signatures of signed commits. If an encountered certificate is not yet trusted, a pop-up will be shown to ask the user whether to trust said certificate. We should disable this by setting the `--no-show-signature` flag. This flag tells git to not show the signature and, in turn, keeps it from trying to verify its validity.

zyzzyxdonta:
This pull request fixes #216


led02 (approved):
Thanks! 🍭

Issue #204, opened 2023-08-11
hermes harvest return 0 on error


daniel-mohr:
Following [Set up automatic software publishing](https://docs.software-metadata.pub/en/latest/tutorials/automated-publication-with-ci.html) I created a `CITATION.cff` and in the end in the pipeline the upload to zenodo sandbox fails with `Error in 'hermes.deposit.publish' entry point 'invenio': With access_right open, a license is required.`.

The reason was that `hermes harvest` already claims:

```
!!! warning "CITATION.cff is not valid according to <https://citation-file-format.github.io/1.2.0/schema.json>"
    Invalid input for `root`.
    !!! message "'message' is a required property"
```

But still the command returned 0 and the pipeline passed. And for a user no error is  recognizable.

I would suggest that `hermes` return an other value than 0 on error. Then the pipeline would break and the user is pointed to the error.

daniel-mohr:
I do not think that a harvesters should be allowed to fail. If the harvester if configured it should work. Otherwise a user can decide to use only some, e. g.:

```
[harvest]
from = [ "git" ]
```

daniel-mohr:
Maybe it could also be interesting to provide the user a simple solution to check the cff file, e. g. as gitlab ci pipeline using [cffconvert](https://github.com/citation-file-format/cffconvert) and an official docker image (The provided [citationcff/cffconvert](https://hub.docker.com/r/citationcff/cffconvert) docker image is pretty old and based on an old alpine:3.14.1.):

```yaml
cffconvert__alpine:
  stage: pre
  image:
    name: alpine:latest
  script:
    - date
    - cat /etc/os-release
    - apk add --no-cache py3-attrs py3-click py3-dateutil py3-docopt py3-jsonschema py3-packaging py3-parsing py3-pip py3-pykwalify py3-pyrsistent py3-requests py3-retrying py3-setuptools py3-six
    - pip3 install cffconvert
    - cffconvert --validate
```

Or using the [pre-commit hook](https://github.com/citation-file-format/cffconvert#pre-commit-hook).

This could be mentioned in the documentation.

led02:
In #253 there is a first fix to this. I.e., `hermes` exits with error code 1 when some command has an error (that might not be fatal), it stops with code 2 when there is an uncaught exception, otherwise the code will be 0.

This is behavior is still subject to change, follow up on issue #238

PR #218, opened 2023-10-10
Class-based architecture for deposition plugins

zyzzyxdonta:
This pull request converts the existing deposition plugins into classes which makes reuse and customization easier and prevents functions from having to load context and config over and over again.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -28,10 +28,9 @@ sys.path.insert(0, os.path.abspath('../../src'))
 
 project = 'HERMES Workflow'
 copyright = '2022, HERMES project'
-author = (
-    'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, Oliver Knodel, '
-    'Michael Meinel, Tobias Schlauch'
-)
+author = 'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, ' + \
+         'Oliver Knodel, Michael Meinel, Tobias Schlauch'
+
 
 # The full version, including alpha/beta/rc tags
 release = '2022-07-01'
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -29,7 +29,7 @@ sys.path.insert(0, os.path.abspath('../../src'))
 project = 'HERMES Workflow'
 copyright = '2022, HERMES project'
 author = 'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, ' + \
-         'Oliver Knodel, Michael Meinel, Tobias Schlauch'
+         'Oliver Knodel, Michael Meinel, Tobias Schlauch, Sophie Kernchen'
 
 
 # The full version, including alpha/beta/rc tags
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -28,7 +28,10 @@ sys.path.insert(0, os.path.abspath('../../src'))
 
 project = 'HERMES Workflow'
 copyright = '2022, HERMES project'
-author = 'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, Oliver Knodel, Michael Meinel, Tobias Schlauch'
+author = (
+    'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, Oliver Knodel, '
+    'Michael Meinel, Tobias Schlauch'
+)
 
 # The full version, including alpha/beta/rc tags
 release = '2022-07-01'
--- a/src/hermes/commands/deposit/base.py
+++ b/src/hermes/commands/deposit/base.py
@@ -16,12 +16,14 @@ class BaseDepositPlugin:
         self.ctx = ctx
 
     def __call__(self) -> None:
-        # TODO: Decide here which of initial/new/... to run. Contract for decision:
-        # Implement method is_new_publication() or similar
         self.prepare()
         self.map_metadata()
-        self.create_initial_version()
-        self.create_new_version()
+
+        if self.is_initial_publication():
+            self.create_initial_version()
+        else:
+            self.create_new_version()
+
         self.update_metadata()
         self.delete_artifacts()
         self.upload_artifacts()
@@ -32,6 +34,9 @@ class BaseDepositPlugin:
 
     def map_metadata(self) -> None:
         pass
+
+    def is_initial_publication(self) -> bool:
+        return True
 
     def create_initial_version(self) -> None:
         pass
--- a/src/hermes/commands/deposit/base.py
+++ b/src/hermes/commands/deposit/base.py
@@ -17,19 +17,14 @@ class BaseDepositPlugin:
 
     def __call__(self) -> None:
         # TODO: Decide here which of initial/new/... to run?
-        steps = [
-            "prepare",
-            "map_metadata",
-            "create_initial_version",
-            "create_new_version",
-            "update_metadata",
-            "delete_artifacts",
-            "upload_artifacts",
-            "publish",
-        ]
-
-        for step in steps:
-            getattr(self, step)()
+        self.prepare()
+        self.map_metadata()
+        self.create_initial_version()
+        self.create_new_version()
+        self.update_metadata()
+        self.delete_artifacts()
+        self.upload_artifacts()
+        self.publish()
 
     def prepare(self) -> None:
         pass
--- a/src/hermes/commands/deposit/base.py
+++ b/src/hermes/commands/deposit/base.py
@@ -1,7 +1,10 @@
 # SPDX-FileCopyrightText: 2023 Helmholtz-Zentrum Dresden-Rossendorf (HZDR)
 #
 # SPDX-License-Identifier: Apache-2.0
+
 # SPDX-FileContributor: David Pape
+# SPDX-FileContributor: Michael Meinel
+
 import abc
 
 import click
--- a/src/hermes/commands/deposit/base.py
+++ b/src/hermes/commands/deposit/base.py
@@ -19,7 +19,7 @@ class BaseDepositPlugin:
         # TODO: Decide here which of initial/new/... to run?
         steps = [
             "prepare",
-            "map",
+            "map_metadata",
             "create_initial_version",
             "create_new_version",
             "update_metadata",
@@ -34,7 +34,7 @@ class BaseDepositPlugin:
     def prepare(self) -> None:
         pass
 
-    def map(self) -> None:
+    def map_metadata(self) -> None:
         pass
 
     def create_initial_version(self) -> None:
--- a/src/hermes/commands/deposit/base.py
+++ b/src/hermes/commands/deposit/base.py
@@ -15,6 +15,10 @@ class BaseDepositPlugin:
         self.ctx = ctx
 
     def __call__(self) -> None:
+        """Initiate the deposition process.
+
+        This calls a list of additional methods on the class, none of which need to be implemented.
+        """
         self.prepare()
         self.map_metadata()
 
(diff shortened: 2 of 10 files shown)


src/hermes/commands/deposit/base.py
```diff
@@ -0,0 +1,56 @@
...
+            "create_initial_version",
+            "create_new_version",
+            "update_metadata",
+            "delete_artifacts",
+            "upload_artifacts",
+            "publish",
+        ]
+
+        for step in steps:
+            getattr(self, step)()
```
led02, line 32: **issue:** It does not feel right to handle it this way... lots of unintended stuff could happen here, some magic might not work. Also all the functions are state-dependant which in my eyes is not the best solution here.

src/hermes/commands/workflow.py
```diff
@@ -189,65 +190,35 @@ def deposit(click_ctx: click.Context, initial, auth_token, file):
...
             # each platform, only a single implementation should exist. Otherwise we
             # would not be able to decide which implementation to choose.
             _log.error(
-                f"Entry point name {name!r} is not unique within group {group!r}"
+                f"Entry point name {plugin!r} is not unique within group {group!r}"
             )
             click_ctx.exit(1)
+    except ValueError:  # not enough values to unpack
+        _log.error(f"Entry point name {plugin!r} not found in group {group!r}")
+        click_ctx.exit(1)
```
led02, line 211: **praise:** That's pythonic, dude 👍

src/hermes/commands/workflow.py
```diff
@@ -189,65 +190,35 @@ def deposit(click_ctx: click.Context, initial, auth_token, file):
...
 
-    for entry_point in loaded_entry_points:
-        try:
-            entry_point(click_ctx, ctx)
-        except (RuntimeError, MisconfigurationError) as e:
-            _log.error(f"Error in {group!r} entry point {name!r}: {e}")
-            click_ctx.exit(1)
+    try:
+        deposit_plugin()
+    except (RuntimeError, MisconfigurationError) as e:
```
led02, line 219: **thought:** I know that the initial draft could be mine... but `RuntimeError` should neither be explicitly raised nor handled by the program IIRC.

src/hermes/commands/deposit/file.py
```diff
@@ -7,24 +7,19 @@
...
 
-def dummy_noop(click_ctx: click.Context, ctx: CodeMetaContext):
-    pass
-
-
-def map_metadata(click_ctx: click.Context, ctx: CodeMetaContext):
-    ctx.update(ContextPath.parse('deposit.file'), ctx['codemeta'])
+class FileDepositPlugin(BaseDepositPlugin):
+    def map(self) -> None:
+        self.ctx.update(ContextPath.parse('deposit.file'), self.ctx['codemeta'])
```
led02, line 17: **thought:** This might as well be the default behaviour (i.e., moved up to base class)

src/hermes/commands/deposit/base.py
```diff
@@ -0,0 +1,56 @@
...
+from hermes.model.context import CodeMetaContext
+
+
+# TODO: Abstract base class?
+class BaseDepositPlugin:
+    def __init__(self, click_ctx: click.Context, ctx: CodeMetaContext) -> None:
+        self.click_ctx = click_ctx
+        self.ctx = ctx
+
+    def __call__(self) -> None:
```
led02, line 18: **thought:** "Explicit is better than implicit" - maybe there should be a method `run` instead?

led02 (commented):
Still WIP, but I could not resist to leave some notes already. Thanks for getting this started 😀

led02:
Maybe we should also write an ADR about this... now that HEQATE is not coming, we need to ensure quality ourselves 😭

zyzzyxdonta:
Yup, I also thought about writing an ADR and was going to address it at the next JF 😄

src/hermes/commands/deposit/base.py
zyzzyxdonta: I called it `run` initially but then thought it looked like a step called `run` and wanted to make it more obvious that that's not what it is. Not sure what to call it 🤷🏻‍♂️

src/hermes/commands/deposit/file.py
zyzzyxdonta: I was hesitant to give any of the actual steps default functionality because I didn't want to force people to think about whether (and when) they need to call `super().step()` in their implementation.

src/hermes/commands/deposit/base.py
```diff
@@ -0,0 +1,56 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# SPDX-FileContributor: David Pape
+
+import click
+
+from hermes.model.context import CodeMetaContext
+
+
+class BaseDepositPlugin:
```
sdruskat, line 12: Are there reasons why you didn't make this an ABC?

zyzzyxdonta: ABCs force the inherited class to implement all methods. E.g. in the `FileDepositPlugin`, this would mean 6 or so additional methods that all `pass` because there's nothing meaningful for them to do.

led02: **opinion:** `ABC` are unpythonic

src/hermes/commands/deposit/base.py
led02: Well, that's fine I think. The only point where this can get problematic is when we want to call a base implementation in a derived class due to Python attribute mangling with dunder names.

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
 import requests
 
 from hermes import config
+from hermes.commands.deposit.base import BaseDepositPlugin
 from hermes.commands.deposit.error import DepositionUnauthorizedError
 from hermes.error import MisconfigurationError
 from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
 from hermes.utils import hermes_user_agent
```
led02, line 25: **nitpick:** This constant looks a lot like a variable...

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
-_DEFAULT_COMMUNITIES_API_PATH = "api/communities"
-_DEFAULT_DEPOSITIONS_API_PATH = "api/deposit/depositions"
 
+_log = logging.getLogger("cli.deposit.invenio")
 
-def prepare(click_ctx: click.Context, ctx: CodeMetaContext):
-    """Prepare the deposition on an Invenio-based platform.
 
-    In this function we do the following:
+class InvenioClient(requests.Session):
```
led02, line 31: **praise:** In general, I like how you split up the functionality into Deposition and Client.

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
+        return self.api_paths.get("records", self.DEFAULT_RECORDS_API_PATH)
 
-    If a previous publication exists, this function does nothing, leaving the work for
-    :func:`create_new_version`.
-    """
 
-    invenio_path = ContextPath.parse("deposit.invenio")
-    invenio_ctx = ctx[invenio_path]
-    latest_record_id = invenio_ctx.get("latestRecord", {}).get("id")
+class InvenioResolver:
```
led02, line 100: **hint:** Could be extended by a `resolve_license_url`

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
-    if latest_record_id is not None:
-        # A previous version exists. This means that we need to create a new version in
-        # the next step. Thus, there is nothing to do here.
-        return
+    invenio_client_class = InvenioClient
 
-    if not click_ctx.params['initial']:
-        raise RuntimeError("Please use `--initial` to make an initial deposition.")
+    def __init__(self, client=None):
+        self.client = client or self.invenio_client_class()
```
led02, line 105: **nitpick:** I have a feeling of YAGNI here

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
+        super().__init__(click_ctx, ctx)
 
+        self.invenio_context_path = ContextPath.parse(f"deposit.{self.platform_name}")
+        self.invenio_ctx = None
 
-def delete_artifacts(click_ctx: click.Context, ctx: CodeMetaContext):
-    """Delete existing file artifacts.
+        auth_token = self.click_ctx.params.get("auth_token")
+        if auth_token is None:
+            raise DepositionUnauthorizedError("No auth token given for deposition platform")
```
led02, line 201: **chore:** This also fails if we inject a fully configured client. I think this should not be the case.

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
+        """Create an initial version of a publication."""
 
-    Anyway, the record id will always be used to resolve the latest version.
+        if not self.click_ctx.params['initial']:
+            raise RuntimeError("Please use `--initial` to make an initial deposition.")
 
-    If any of the resolution steps fail or produce an unexpected result, a ValueError will be thrown.
+        response = self.client.new_deposit()
+        if not response.ok:
+            raise RuntimeError(f"Could not create initial deposit {response.url!r}")
```
led02, line 277: **chore:** The answer returned by the server is very important here and should be either logged or be part of the exception message.

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
+        response = self.client.put(
+            draft_url,
+            json={"metadata": deposition_metadata}
         )
 
-    parsed_url = urlparse(license_url)
-    url_path = parsed_url.path.rstrip("/")
-    license_id = url_path.split("/")[-1]
+        if not response.ok:
+            raise RuntimeError(f"Could not update metadata of draft {draft_url!r}")
```
led02, line 318: **chore:** Again, we need to know about the error message returned by the server.

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
-        url = f"{communities_api_url}/{community_id}"
-        response = session.get(url)
-        if response.status_code == 404:
-            raise MisconfigurationError(
-                f"Not a valid community identifier: {community_id}"
-            )
-        # Catch other problems
-        response.raise_for_status()
-        community_ids.append({"identifier": response.json()["id"]})
+        files: list[click.Path] = self.click_ctx.params["file"]
```
led02, line 346: **question:** Under which conditions do we add type hints for local variables?

src/hermes/commands/deposit/invenio.py
```diff
@@ -17,648 +17,613 @@
...
+            "references": None,
+            "communities": communities,
+            "grants": None,
+            "subjects": None,
+            "version": metadata.get('version'),
+        }.items() if v is not None}
+
+        return deposition_metadata
+
+    def _get_license_identifier(self):
```
led02, line 496: **question:** Why is this not part of the *Resolver*?

docs/source/conf.py
```diff
@@ -28,7 +28,10 @@
 
 project = 'HERMES Workflow'
 copyright = '2022, HERMES project'
-author = 'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, Oliver Knodel, Michael Meinel, Tobias Schlauch'
+author = (
+    'Oliver Bertuch, Stephan Druskat, Guido Juckeland, Jeffrey Kelling, Oliver Knodel, '
+    'Michael Meinel, Tobias Schlauch'
+)
```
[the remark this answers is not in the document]
led02: **crazy idea:** Maybe, we could also get this information from ... hermes 😮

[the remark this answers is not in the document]
led02: ... also, what is the correct order?

src/hermes/commands/deposit/invenio.py
led02: ... but looking into the code, it could be just enough of dynamic involved to regard it as a variable.

src/hermes/commands/workflow.py
led02: **todo:** Make this an issue (i.e., general strategy for error handling) ... my guts says, there was something like this already but I could not find a matching issue. At least I'm sure we spoke about it already (i.e., that we need to do error handling in a consistent way)
This is also related to our community contribution #204

src/hermes/commands/deposit/base.py
```diff
@@ -0,0 +1,81 @@
...
```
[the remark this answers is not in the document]
led02: **todo:** This needs to be addressed in an ADR and also needs thorough documentation (as soon as it is fixed...)

led02 (approved):

src/hermes/commands/deposit/base.py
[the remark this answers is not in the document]
led02: This is handled in #192

PR #222, opened 2023-10-18
(Try to) Fix InvenioRDM license checks

### LINKED ISSUE #221: Fix Invenio deposition URLs
Zenodo switched to InvenioRDM 🎉 

Zenodo did not update the API docs yet, but some paths changed...

- api/licenses --> api/vocabularies/licenses (see also #222, #224)

### LINKED ISSUE #224: Invenio RDM license identifiers are lowercase, while spdx links are not, causing deposit to fail
Invenio RDM exposes the licenses ids in lowercase in the API (At least for zenodo and zenodo sandbox):

i.e  For the MIT license:
valid endpoint:

```https://sandbox.zenodo.org/api/vocabularies/licenses/mit```
Non valid endpoint
```https://sandbox.zenodo.org/api/vocabularies/licenses/MIT```

Since all SPDX links have Uppercase or mixed, but not lowercase (https://spdx.org/licenses/MIT) 
This code part https://github.com/hermes-hmc/workflow/blob/5726d30e9b11a161034588f3c897debb7ef8f968/src/hermes/commands/deposit/invenio.py#L556-L560
will fail. i.e the license id probably has to be converted to lowercase. I am currently not sure if this will work for all identifiers.
Or better, one first has to extract an 'identifier mapping' from an API endpoint.

This currently causes Hermes deposit with:
```
Error in 'hermes.deposit.publish' entry point 'invenio': Not a valid license identifier: MIT
```

Hermes config:

```
[harvest]
from = [ "git", "cff" ]

[harvest.cff]
validate = false

[deposit]
mapping = "invenio"
target = "invenio"

[deposit.invenio]
site_url = "https://sandbox.zenodo.org"
access_right = "open"

[deposit.invenio.api_paths]
depositions = "api/deposit/depositions"
licenses = "api/vocabularies/licenses"
communities = "api/communities"

[postprocess]
execute = [ "config_record_id" ]

```

### PR DESCRIPTION
**Pull request by led02**
Fixes #221
Fixes #224

### DIFF
--- a/hermes.toml
+++ b/hermes.toml
@@ -9,15 +9,15 @@ from = [ "cff", "git" ]
 validate = false
 
 [deposit]
-mapping = "invenio"
-target = "invenio"
+mapping = "invenio_rdm"
+target = "invenio_rdm"
 
-[deposit.invenio]
+[deposit.invenio_rdm]
 site_url = "https://sandbox.zenodo.org"
 communities = ["zenodo"]
 access_right = "open"
 
-[deposit.invenio.api_paths]
+[deposit.invenio_rdm.api_paths]
 depositions = "api/deposit/depositions"
-licenses = "api/licenses"
+licenses = "api/vocabularies/licenses"
 communities = "api/communities"
--- a/hermes.toml
+++ b/hermes.toml
@@ -9,7 +9,6 @@ from = [ "cff", "git" ]
 validate = false
 
 [deposit]
-mapping = "invenio_rdm"
 target = "invenio_rdm"
 
 [deposit.invenio_rdm]
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -23,7 +23,7 @@ from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
 from hermes.utils import hermes_user_agent
 
-_DEFAULT_LICENSES_API_PATH = "api/vocabulary/licenses"
+_DEFAULT_LICENSES_API_PATH = "api/vocabularies/licenses"
 _DEFAULT_COMMUNITIES_API_PATH = "api/communities"
 _DEFAULT_DEPOSITIONS_API_PATH = "api/deposit/depositions"
 
@@ -553,7 +553,7 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
 
     # Fetch full list of licenses available... maybe we should cache this.
     response = requests.get(
-        f"{license_api_url}", headers={"User-Agent": hermes_user_agent}
+        f"{license_api_url}?size=1000", headers={"User-Agent": hermes_user_agent}
     )
     response.raise_for_status()
 
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -555,6 +555,8 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
     url_path = parsed_url.path.rstrip("/")
     license_id = url_path.split("/")[-1].lower()
 
+    print(f"DEBUG: License is said to be {license_url} -> {license_id}")
+    
     response = requests.get(
         f"{license_api_url}/{license_id}", headers={"User-Agent": hermes_user_agent}
     )
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -553,7 +553,7 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
 
     parsed_url = urlparse(license_url)
     url_path = parsed_url.path.rstrip("/")
-    license_id = url_path.split("/")[-1]
+    license_id = url_path.split("/")[-1].lower()
 
     response = requests.get(
         f"{license_api_url}/{license_id}", headers={"User-Agent": hermes_user_agent}
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -23,7 +23,7 @@ from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
 from hermes.utils import hermes_user_agent
 
-_DEFAULT_LICENSES_API_PATH = "api/licenses"
+_DEFAULT_LICENSES_API_PATH = "api/vocabulary/licenses"
 _DEFAULT_COMMUNITIES_API_PATH = "api/communities"
 _DEFAULT_DEPOSITIONS_API_PATH = "api/deposit/depositions"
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/hermes/commands/deposit/invenio.py [hunk 1]
```diff
@@ -553,8 +553,10 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
 
     parsed_url = urlparse(license_url)
     url_path = parsed_url.path.rstrip("/")
-    license_id = url_path.split("/")[-1]
+    license_id = url_path.split("/")[-1].lower()
 
+    print(f"DEBUG: License is said to be {license_url} -> {license_id}")
+
```
led02, line 559: **todo:** When did we start to use `printf` debugging in HERMES? This is just a joke I don't get, right?

src/hermes/commands/deposit/invenio.py
```diff
@@ -551,19 +551,39 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
...
+    def _search_license_info(_url):
+        for license_info in valid_licenses['hits']['hits']:
+            try:
+                if license_info['props']['url'] == _url:
+                    return license_info
+            except KeyError:
+                continue
+        else:
+            return None
```
led02, line 569: **question:** Do we like nested functions?

src/hermes/commands/deposit/invenio.py
```diff
@@ -551,19 +551,39 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
...
+
+        for license_cross_ref in response.json()['crossRef']:
+            if not license_cross_ref['isValid']:
+                continue
+
+            license_info = _search_license_info(license_cross_ref["url"])
+            if license_info is not None:
+                break
+        else:
+            raise RuntimeError(f"Could not resolve license URL {license_url} to a valid identifier.")
```
led02, line 584: **hint:** Yes, this is very ugly, but seems to be robust to me, even when InvenioRDM decides to use poper SPDX URLs.

src/hermes/commands/deposit/invenio.py
```diff
@@ -551,19 +551,39 @@ def _get_license_identifier(ctx: CodeMetaContext, license_api_url: str):
...
         )
 
-    parsed_url = urlparse(license_url)
-    url_path = parsed_url.path.rstrip("/")
-    license_id = url_path.split("/")[-1]
-
+    # Fetch full list of licenses available... maybe we should cache this.
     response = requests.get(
-        f"{license_api_url}/{license_id}", headers={"User-Agent": hermes_user_agent}
+        f"{license_api_url}?size=1000", headers={"User-Agent": hermes_user_agent}
```
zyzzyxdonta, line 556: I haven't looked at this PR yet. Just one quick thing: It might be worth it to check `f"{license_api_url}/{license_id.lower()}"` first. before downloading all the licenses. See here: https://zenodo.org/api/vocabularies/licenses/apache-2.0

led02: Good point.

led02: Added in [2364bc3](https://github.com/hermes-hmc/workflow/pull/222/commits/2364bc34804776226557f06fce619be52b5145b3)

zyzzyxdonta:
Question: Should we keep the Invenio implementation intact and make this the start of a new InvenioRDM implementation that then moves towards [the new InvenioRDM API](https://inveniordm.docs.cern.ch/reference/rest_api_index/)?

led02:
> Question: Should we keep the Invenio implementation intact and make this the start of a new InvenioRDM plugin that then moves towards [the new InvenioRDM API](https://inveniordm.docs.cern.ch/reference/rest_api_index/)?

**IMHO:** Well, we could keep the old plugin intact "for historical reasons" but I don't see an actual reason for it as it would be no longer used. However, a good argument for keeping the old one is to signal our numerous users that there was a change (since we do not version our plugins (yet?)).

So to conclude: Yeah, let's do an InvenioRDM plugin.

zyzzyxdonta:
>I don't see an actual reason for it as it would be no longer used

Zenodo was not the only Invenio instance out there 😉

hermes.toml
```diff
@@ -9,15 +9,15 @@ from = [ "cff", "git" ]
 validate = false
 
 [deposit]
-mapping = "invenio"
-target = "invenio"
+mapping = "invenio_rdm"
```
zyzzyxdonta, line 12:
```suggestion
```

The "mapping" setting is not used any more.

src/hermes/commands/deposit/invenio_rdm.py
```diff
@@ -0,0 +1,717 @@
...
+    of type :class:`str`, a :class:`RuntimeError` is raised.
+
+    Invenio instances take a license string which refers to a license identifier.
+    Typically, Invenio instances offer licenses from https://opendefinition.org and
+    https://spdx.org. However, it is possible to mint PIDs for custom licenses.
+
+    An API endpoint (usually ``/api/vocabularies/licenses``) can be used to check whether a given
+    license is supported by the Invenio instance. This function tries to retrieve the
+    license by the identifier at the end of the license URL path. If this identifier
+    does not exist on the Invenio instance, a :class:`RuntimeError` is raised. If no
```
zyzzyxdonta, line 540: This needs updating. Now, other things are checked before giving up.

src/hermes/commands/deposit/invenio_rdm.py
```diff
@@ -0,0 +1,717 @@
...
+                if license_info['props']['url'] == _url:
+                    return license_info
+            except KeyError:
+                continue
+        else:
+            return None
+
+    license_info = _search_license_info(license_url)
+    if license_info is None and license_url.startswith('https://spdx.org/licenses/'):
+        response = requests.get(f"{license_url}.json", headers={"User-Agent": hermes_user_agent})
```
zyzzyxdonta, line 606: Content negotiation would be nice. But they don't seem to support it 🙄 😅

src/hermes/commands/deposit/invenio_rdm.py
```diff
@@ -0,0 +1,717 @@
...
+        deposit_url,
+        json={"metadata": deposition_metadata},
+        headers={
+            "User-Agent": hermes_user_agent,
+            "Authorization": f"Bearer {click_ctx.params['auth_token']}",
+        }
+    )
+
+    if not response.ok:
+        print(response.text)
```
zyzzyxdonta, line 137: Add this to the `RuntimeError` or use logging instead?

zyzzyxdonta (changes requested):
Just some minor nitpicks 😄

src/hermes/commands/deposit/invenio_rdm.py
led02: **joke:** And I though we only want to avoid *print**f*** debugging...

src/hermes/commands/deposit/invenio_rdm.py
led02: Yes, unfortunate they use the different URLs. However, it's at least documented... even though I could not find the source 😅

src/hermes/commands/deposit/invenio_rdm.py
led02: Yes, unfortunate they use the different URLs. However, it's at least documented... even though I could not find the source 😅

led02:
**nitpick:** https://conventionalcomments.org states on *nickpick*

> Nitpicks are trivial preference-based requests. These should be non-blocking by nature.

zyzzyxdonta:
> **nitpick:** https://conventionalcomments.org states on _nickpick_
> 
> > Nitpicks are trivial preference-based requests. These should be non-blocking by nature.
> 
> However, all your requested changes were well founded and I was happy to fix them. So maybe they were more _todo_s?

😂 Good point

zyzzyxdonta (approved):

Issue #201, opened 2023-08-10
Tutorial: Gitlab HERMES_PUSH_TOKEN not documented


daniel-mohr:
[Set up automatic software publishing](https://docs.software-metadata.pub/en/latest/tutorials/automated-publication-with-ci.html) describes how to set up your workflow.

But it is missing to mention that one has to provide the CI variable `HERMES_PUSH_TOKEN`. This token has to have at least the role `Developer` and needs write access to the project to create branches.

led02:
This relates to the GitLab example / template.

daniel-mohr:
Ah, yes. I used your gitlab workflow/example/template.

aghr:
I ran into the same issue and solved it myself after online investigation. Please update your documentation as soon as possible to avoid that more GitLab users run into this issue. Good documentation is vital for your users. Many thanks.

Issue #203, opened 2023-08-10
authentication in non protected branch


daniel-mohr:
[Set up automatic software publishing](https://docs.software-metadata.pub/en/latest/tutorials/automated-publication-with-ci.html) describes how to set up your workflow.

First it creates some branches and a merge request from `hermes/curate-result-...` into `hermes/curate-...`.

Accepting this merge request runs the pipeline in the branch `hermes/curate-...`. This branch is of cause not protected and therefore it has no access to protected variables.

But in [hermes-ci.yml](https://github.com/hermes-hmc/ci-templates/blob/993db221ea3319cddfd65e7215954eed1bea9b10/gitlab/hermes-ci.yml#L58) the `HERMES_PUSH_TOKEN` (cf. #201) is necessary. Unfortunately the `HERMES_PUSH_TOKEN`  stores a secret and is therefore only available in masked branches.

daniel-mohr:
Making the branch `hermes/curate-...` protected does not help. This breaks with `remote: GitLab: You are not allowed to delete protected branches from this project. Only a project maintainer or owner can delete a protected branch.`.

led02:
This relates to the GitLab example / template.

daniel-mohr:
Yes, I used your gitlab workflow/example/template.

daniel-mohr:
Using persitent branches could allow to setup a gitlab project to allow the `HERMES_PUSH_TOKEN` to work in these branches.

This would be the same workflow like [GitHub Pages Deploy Action](https://github.com/JamesIves/github-pages-deploy-action) which also stores the documentation in a persitent branch which is then used by [github pages](https://pages.github.com/).

daniel-mohr:
If I remember correctly [GitHub Pages Deploy Action](https://github.com/JamesIves/github-pages-deploy-action) uses orphan branches.

Using orphan branches would also prevent merging the hermes branches.

daniel-mohr:
As discussed with @led02 I believe your used workflow has a few drawbacks and could be enhanced. I am talking here from using hermes with gitlab (In general it should be similar in github; but zenodo has a github integration which makes it much easier.).

I see a few possibilities:

### 1. curating merge request with transient branches (your current workflow)

[...]

PR #223, opened 2023-10-18
Add HERMES_PUSH_TOKEN

SKernchen:
See #201 for Information.

### DIFF
--- a/docs/source/tutorials/automated-publication-with-ci.md
+++ b/docs/source/tutorials/automated-publication-with-ci.md
@@ -8,6 +8,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 SPDX-FileContributor: Oliver Bertuch
 SPDX-FileContributor: Michael Meinel
 SPDX-FileContributor: Stephan Druskat
+SPDX-FileContributor: Sophie Kernchen
 -->
 
 # Set up automatic software publishing
--- a/docs/source/tutorials/automated-publication-with-ci.md
+++ b/docs/source/tutorials/automated-publication-with-ci.md
@@ -172,6 +172,13 @@ Therefore, you [create an access token in your project](https://docs.gitlab.com/
 The Token needs to have at least the `developer` role and `write` access (e.g. write_repository scope).
 Then you create a Gitlab CI variable with the token called `HERMES_PUSH_TOKEN`.
 
+```{note}
+The two Gitlab CI Variables include sensitive and powerful information.
+Therefore you should at least select the flag `Mask variable` when creating.
+If possible you should also select the flag `Protect variable` and define all branches `hermes/*` as
+protected branch.
+```
+
 Copy the [template file for GitLab to Zenodo Sandbox publication](https://github.com/hermes-hmc/ci-templates/blob/main/gitlab/hermes-ci.yml)
 into your project to `.gitlab/hermes-ci.yml`.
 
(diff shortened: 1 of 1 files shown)


daniel-mohr:
Since the necessary `HERMES_PUSH_TOKEN` is really powerful, I would also suggest to warn/inform the user, e. g.:

```md
The `HERMES_PUSH_TOKEN` is really powerful! Therefore you should ensure that only people with at least
the same rights on your project could access this CI variable.
This means you should at least select the flag `Mask variable`.
If possible you should also select the flag `Protect variable` and define all branches `hermes/*` as
protected branch.
```

Further I would suggest a similar warning regarding the `ZENODO_TOKEN`, e. g.:

```md
The `ZENODO_TOKEN` is a personal access token to your invendio/zenodo account.
Therefore you should ensure that you do not share it with other people.
This means you have to be the only `Developer`, `Maintainer` and `Owner` in your project.
Further you should at least select the flag `Mask variable`.
If possible you should also select the flag `Protect variable` and define all branches `hermes/*` as
protected branch.
```

SKernchen:
> Since the necessary `HERMES_PUSH_TOKEN` is really powerful, I would also suggest to warn/inform the user, e. g.:
> 
> ```
> The `HERMES_PUSH_TOKEN` is really powerful! Therefore you should ensure that only people with at least
> the same rights on your project could access this CI variable.
> This means you should at least select the flag `Mask variable`.
> If possible you should also select the flag `Protect variable` and define all branches `hermes/*` as
> protected branch.
> ```
> 
> Further I would suggest a similar warning regarding the `ZENODO_TOKEN`, e. g.:
> 
> ```
> The `ZENODO_TOKEN` is a personal access token to your invendio/zenodo account.
> Therefore you should ensure that you do not share it with other people.
> This means you have to be the only `Developer`, `Maintainer` and `Owner` in your project.
> Further you should at least select the flag `Mask variable`.
> If possible you should also select the flag `Protect variable` and define all branches `hermes/*` as
> protected branch.
> ```

Thank you for contributing @daniel-mohr , i will adapt your remarks 😃.

daniel-mohr:
@SKernchen I think you missed my inline comment / review comment, isn' it?

cf.:
> `write_repository` (write access to the repository) is enough
> `api` (write access to the project)

SKernchen:
@daniel-mohr Oh I'm sorry but what exactly do you mean?
For the Hermes Push Token write acces is needed.
In the tutorial i propose to use the `write_repository` scope which is the more limited than the api scope.
There is also the documentation link for creating personal access token where one individual can see which fits best.

daniel-mohr:
@SKernchen I repeat here my complete [inliine comment / review comment](https://github.com/hermes-hmc/workflow/pull/223/files):

Maybe it would help to use the gitlab vocabulary:

* `developer` -> `Developer`

Further I do not think `write_repository` (write access to the repository) is enough. At the actual workflow -- as I used it a few months ago -- creates branches. In this way the necessary scope would be `api` (write access to the project).

In #203 I proposed some other possible workflows. Some of these would allow to use `write_repository` only.

daniel-mohr:
So, the Hermes Push Token needs write access to the project. It is not enough to have write access to the repository!

led02:
> So, the Hermes Push Token needs write access to the project. It is not enough to have write access to the repository!

Well, we're just in process of integrating this into a GitLab training and `write_repository` is indeed enough

daniel-mohr:
@led02 But the workflow creates still branches, isn't like this anymore?

led02:
> @led02 But the workflow creates still branches, isn't like this anymore?

Yes, it does create branches and a merge request. But it does so by using Git push magic:
```
git push origin "$MR_SOURCE_BRANCH"
    -o merge_request.create
    -o merge_request.target="$MR_TARGET_BRANCH"
    -o merge_request.title="$MR_TITLE"
    -o merge_request.description="$MR_DESCRIPTION"
    -o merge_request.remove_source_branch
```

This does not require API access.

daniel-mohr:
@SKernchen Thanks for solving my issue!

Issue #156, opened 2023-04-19
Reverse precedence of configured harverster results during merge


jkelling:
At the moment successive merge steps belonging to successive harvester runs override previous values on pre-existing keys.

Harvesters are and their merge steps are executed in the order they are listed in the harvest section of the configuration, leading the last one to take precedence. This seems counter-intuitive, especially considering the current default/example [hermes.toml](hermes.toml):
https://github.com/hermes-hmc/workflow/blob/05e4e530af77bd38810dd802c5c4f1b8249a8ca5/hermes.toml#L5-L6
where a value produce by the `git` harvester,  which could only be a commit hash, would override a value read from a CFF file. #154 does not change this behavior apart from storing the previous value as an alternative.

The are two general ways to address this: either
1. run the harvesters in reverse order, or
2. invert the behavior the the merger to not touch existing values (and add new values as alternatives in #154 ).

Note, that the second option cannot left for the individual harvesters to implement, because
1. this would make implementing a merge strategy mandatory for each harvester, to the detriment of the harvester developers, and
2. would allow for inconsistent harvester precedence, to the detriment of the user.

led02:
In principal, this behaviour could be easily implemented in the current code (at least to *flip the precedence*).

Do we want that already in Proof-of-Concept? We should definately re-consider it when discussing the breaking changes for Alpha.

SKernchen:
The order is flipped so the first harvester source is the most important.

Issue #182, opened 2023-05-22
Rename --path to -C,--directory


jkelling:
Maybe this option should be named more in line with established conventions.

I just did a quick survey and both make and poetry use `-C, --directory` for the CLI option with this semantics, so I suggest to use the same.

_Originally posted by @jkelling in https://github.com/hermes-hmc/workflow/pull/181#discussion_r1200069566_

poikilotherm:
This is related, a subset or even a duplicate of #137 and #179.

led02:
> Maybe this option should be named more in line with established conventions.

Where are these conventions defined? Why should we comply to them? Or is this simply a convention which is due to the set of tools you usually use?

Don't get me wrong: If there really is a common convention that holds for the majority of tools, we should follow it. However, I could not find a single reference to what flags should be named like (but only to how the syntax should be).

PR #231, opened 2023-11-15
Replace click with argparse

### LINKED ISSUE #137: Straighten out CLI usage

sdruskat:
This PR removes click from the direct dependencies of hermes, and re-implements the CLI functionality with `argparse`. Multi-commands a.k.a. subcommand chaining is removed.

- Fixes #137 

- [ ] Supersede ADR10 with argparse ADR and no multi-commands


led02:
**question:** While working on this, should I also handle #182 (which will break the API and will require adaptations in docs @SKernchen and probably workflows @poikilotherm)

led02:
**swear:** Obviously, somebody ran an auto formatter on this branch, forcing me to manually merge tons of commits with uncountable lines where only the quotation marks were changed. But hidden in between are changes that come from the Pydantic Settings change. 😡

src/hermes/commands/workflow.py
```diff
@@ -207,27 +221,28 @@ def deposit(click_ctx: click.Context, initial, auth_token, file):
...
-        deposit_plugin()
-    except (RuntimeError, MisconfigurationError) as e:
-        _log.error(f"Error in {plugin_group!r} plugin {plugin_name!r}: {e}")
-        click_ctx.exit(1)
+    for entry_point in loaded_entry_points:
+        try:
+            deposit_plugin(path, config_path, initial, auth_token, files, ctx)
+        except (RuntimeError, MisconfigurationError) as e:
+            _log.error(f"Error in {group!r} entry point {name!r}: {e}")
+            sys.exit(1)
```
zyzzyxdonta, line 237: There's something wrong here. `loaded_entry_points` does not exist and neither do `group` and `name`.

led02:
> I like where this is going 👍🏻 However, it doesn't seem to be wired up correctly yet, as none of the commands do anything so far.

Focus was on the cli.py until now. The design of the command line interface should be kind of stable, already started to do some magic Pydantic Settings stuff here. Unfortunatly, SWIM ran a code formatter and pushed the results, hence there are some more changes indicated that there really are... :(

Still need to:

- wire up the implementation
- finish the dynamic settings assembly
- make the other phases also use plug-in classes

src/hermes/commands/workflow.py
```diff
@@ -7,13 +7,14 @@
 # SPDX-FileContributor: David Pape
 # SPDX-FileContributor: Oliver Bertuch
 
+import argparse
```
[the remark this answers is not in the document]
led02: The `workflow.py` will be gone once this PR is ready.

src/hermes/commands/deposit/base.py
```diff
@@ -21,8 +19,7 @@ class BaseDepositPlugin(HermesPlugin):
     TODO: describe workflow... needs refactoring to be less stateful!
     """
 
-    def __init__(self, click_ctx: click.Context, ctx: CodeMetaContext) -> None:
-        self.click_ctx = click_ctx
+    def __init__(self, ctx):
         self.ctx = ctx
```
led02, line 23: **thought:** We might decide to keep the ``HermesContext`` and derived mischief for now. I see the need to refactor the "deposit" (base) plugin which is heavily state dependent. However, we are doing data processing, hence I think it's a good idea to reflect that in the design...

However, to go forward I think it's easiest to keep the ctx stuff for now, design the plugin interface in a way that only the "Command" does interact with it and have a better hinge point for later refactoring.

To sum it up: For now, do the refactoring using context, design new plugins that the data is visible, later get rid of contexts.
Disclaimer: The purpose of this note is to derive an issue out of it ;)

src/hermes/commands/deposit/invenio.py
```diff
@@ -252,14 +252,14 @@ class InvenioDepositPlugin(BaseDepositPlugin):
...
-        super().__init__(click_ctx, ctx)
+    def __init__(self, ctx: CodeMetaContext, client=None, resolver=None) -> None:
+        super().__init__(ctx)
 
         self.invenio_context_path = ContextPath.parse(f"deposit.{self.platform_name}")
         self.invenio_ctx = None
 
         if client is None:
-            auth_token = self.click_ctx.params.get("auth_token")
+            auth_token = self.ctx.params.get("auth_token")
```
led02, line 262: **chore:** We need to decide on how to pass command line arguments to the plugins. Currently, the only thing that plugins can interact with the "outside" world is the settings. Hence, I'd opt for putting everything that is needed by the plugin in to a setting. You can also configure those settings from command line by using e.g,, ``hermes harvest -O invenio.auth_tokten MY_AUTH_TOKEN -O invenio.initial True``

src/hermes/commands/process/base.py
```diff
@@ -17,6 +21,11 @@ class HermesProcessCommand(HermesCommand):
     """ Process the collected metadata into a common dataset. """
 
     command_name = "process"
+    settings_class = None
 
     def __call__(self, args: argparse.Namespace) -> None:
-        pass
+        self.args = args
+        # TODO: get harvested data
+        # TODO: Merge Datasets
```
led02, line 29: Also:
- Call pre-processors plugins (well, not really clear)
- Call processor plugins

These interfaces are not at all well-defined by now. We had different entry points, but I don't see this suitable for the new CLI without introducing new commands.

I cloud image that **processor** plugins have a pre-process stage (that is called before the merge) and a process stage (that is called after the merge). As we do not have lots to process for now, it's also fine to work without plugins here for now.

SKernchen:
- Fixes #156 
The processor will prioritize the first harvester in the toml.

PR #232, opened 2023-11-15
Update ADRs about serialization of provenance data

### LINKED ISSUE #230: Update ADR 11 for better serialization of provenance data in harvesting and processing
This task is linked to #153 and #127, as it describes the technical implementation.
It is, however, also related to #95 and #22.

To enable better curation and debugging, we need to ship metadata about the metadata we collate. What are sources of the data? Which one was selected (and why)? 

The internal data model ADRs do not yet describe how we implement this. A new ADR is needed.

poikilotherm:
Closes #230

### DIFF
--- a/docs/adr/0002-use-a-common-data-model.md
+++ b/docs/adr/0002-use-a-common-data-model.md
@@ -14,7 +14,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 We need a data model that's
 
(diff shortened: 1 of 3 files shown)


poikilotherm:
Here is the preview of the most important ADR 11: https://hermes-workflow--232.org.readthedocs.build/en/232/adr/0011-record-provenance-of-metadata.html

led02 (approved):
Thanks for the good documentation. I'm fine with this

Issue #219, opened 2023-10-17
Declarative approach to settings


zyzzyxdonta:
We should consider describing the settings for the hermes workflow in a declarative manner, e.g. using [pydantic settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/). This allows us to define the structure of the settings in one (obvious) place rather than throughout the imperative code. It also makes configuration less error-prone as everything is laid out explicitly rather than implicitly and it allows for validation on program startup.

poikilotherm:
Sketch from meeting on 2023-10-25:

- Create a specification class infrastructure to express what kind of settings and/or command line arguments a plugin requires / offers.
- Add a base class method as API contract to retrieve the specifications
- Create a CLI argument parser with static methods that understands the specs and provides lookup
- Create a Config file parser with static methods that understands the specs and provides lookups

PR #234, opened 2023-11-15
Declarative approach to settings

SKernchen:
- #219 
- [ ] Adapt Tutorials


pyproject.toml
```diff
@@ -42,6 +42,7 @@ pyparsing = "^3.0.9"
 requests = "^2.28.1"
 
 # Packages for developers
+pydantic-settings = "^2.1.0"
```
led02, line 45: **nitpick:** Add a blank line for better readability.

```suggestion
pydantic-settings = "^2.1.0"

```

src/hermes/config.py
```diff
@@ -11,6 +11,8 @@
 import toml
 
 from hermes.model.context import HermesContext
+from hermes.settings import HermesSettings
+from pydantic import ValidationError
```
led02, line 15: **todo:** Import of thrid party should appear before internal imports (i.e., just after `import toml`). See [PEP 8](https://peps.python.org/pep-0008/#imports) for details.

led02 (commented):
One minor change.

docs/source/conf.py
```diff
@@ -47,6 +47,7 @@
     'sphinx.ext.ifconfig',
     'sphinx.ext.githubpages',
     'sphinx.ext.autodoc',
+    'sphinx.ext.napoleon',
```
led02, line 50: **thought:** While this fixes the RTD build, it might be a better idea to provide our own docstrings...

src/hermes/config.py
SKernchen, line 15:
```suggestion
import toml
from pydantic import ValidationError

from hermes.model.context import HermesContext
from hermes.settings import HermesSettings
```

zyzzyxdonta (changes requested):

zyzzyxdonta:
Similiarly to the `JsonConfigSettingsSource` defined [here](https://docs.pydantic.dev/latest/concepts/pydantic_settings/), you should be able to write a `TomlConfigSettingsSource` that reads the config from `hermes.toml`.

Then you can instantiate the settings like this:

```python
settings = HermesSettings()
```

at the end of `config.py` and they will be filled with values from the file Then no manual calls to to `model_validate` are needed.

(I'm not sure if we can go this route with the plan of plugins being able to configure their own settings.)

SKernchen:
> Similiarly to the `JsonConfigSettingsSource` defined [here](https://docs.pydantic.dev/latest/concepts/pydantic_settings/), you should be able to write a `TomlConfigSettingsSource` that reads the config from `hermes.toml`.
> 
> Then you can instantiate the settings like this:
> 
> ```python
> settings = HermesSettings()
> ```
> 
> at the end of `config.py` and they will be filled with values from the file Then no manual calls to to `model_validate` are needed.
> 
> (I'm not sure if we can go this route with the plan of plugins being able to configure their own settings.)

I can already skip the model_validate by using the json in the Settings class which uses the HermesSettings. Still defining the Source and automatically reading the Toml could be better for usage.

zyzzyxdonta:
You call `model_validate` [here](https://github.com/hermes-hmc/hermes/pull/234/files#diff-cf27d336265ddc8933b1806a3c3fd1d4f69b8e59a4138dea75860047fa7dbe92R85). Which makes sense because the settings are read manually, validated once and then access using the old code. I.e. `hermes.config.get`. I meant that this could be avoided.

led02:
After reviewing the code, some general remarks here:

**praise:** Worked better than I expected according to your reports. Also your code is well structured and your names seem well chosen. :)

**opinion:** I think it is not required / advisible to try to keep the exiting code working with declarative model.
In contrast, especially the `config.get('foo', 'bar')` calls should be avoided.
Instead, there should be a `HermesSettings` derived class that defines the defaults (where applicable).
While this will bring in lots of minor code breaks, this also brings some major improvements:

- We have a centeral place (for each plugin) where all the required and default values are declared.
- The code to access to the single configuration values is much cleaner.
- We have validation "up front" (i.e., not only when the plugin tries to read the value).
  This is also already given by the validation during `configuration` but emulation of the dictionaries `get` behaviour might lead to unforseen effects.

We could discuss whether it would be good to have both options available for transition but I'd prefer to break and repair instead of maintaining *deprecated* stuff. We could also change your `__getattr__` method into a `get` implementation but this triggers a similar feeling of *wrongness* deep within my programmers ❤️ ... ;)

**hint:** I suppose you are using an automatic code formatter? You should definately extend the allowed line length to 120 chars. This will make the code a bit more readable at some places... (This is also the value we are using [in our liniting](https://github.com/hermes-hmc/hermes/blob/00cd6c45e9315d09aa688db37b93c9b2d4c9246b/.flake8#L7)).

src/hermes/commands/deposit/file.py
```diff
@@ -23,7 +23,7 @@ def map_metadata(click_ctx: click.Context, ctx: CodeMetaContext):
 
 
 def publish(click_ctx: click.Context, ctx: CodeMetaContext):
-    file_config = config.get("deposit").get("file", {})
+    file_config = config.deposit.get("file", {})
```
led02, line 26:
```suggestion
    file_config = config.deposit.file
```

src/hermes/commands/deposit/invenio.py
```diff
@@ -48,7 +48,7 @@ def prepare(click_ctx: click.Context, ctx: CodeMetaContext):
         raise DepositionUnauthorizedError("No auth token given for deposition platform")
 
     invenio_path = ContextPath.parse("deposit.invenio")
-    invenio_config = config.get("deposit").get("invenio", {})
+    invenio_config = config.deposit.get("invenio", {})
```
led02, line 51:
```suggestion
    invenio_config = config.deposit.invenio
```

src/hermes/commands/deposit/invenio.py
```diff
@@ -575,7 +575,7 @@ def _get_community_identifiers(ctx: CodeMetaContext, communities_api_url: str):
     raised.
     """
 
-    communities = config.get("deposit").get("invenio", {}).get("communities")
+    communities = config.deposit.get("invenio", {}).get("communities")
```
led02, line 578:
```suggestion
    communities = config.deposit.invenio.communities
```

src/hermes/config.py
```diff
@@ -76,50 +74,8 @@ def configure(config_path: pathlib.Path, working_path: pathlib.Path):
...
-        _config[name] = _config['hermes'][name]
-
-    elif name != 'hermes' and _config['hermes'][name] is not _config[name]:
-        # If a configuration was loaded, after the defaults were set, update it.
-        _config[name].update(_config['hermes'].get('name', {}))
-
-    return _config.get(name)
-
+    _config['hermes'] = settings
+    _config['logging'] = settings.get('logging', _config['logging'])  # TODO:  'dict' object is not callable
```
led02, line 78:
```suggestion
    _config['logging'] = settings.logging
```

src/hermes/settings.py
```diff
@@ -0,0 +1,121 @@
...
+                field_name, field, field_value, value_is_complex
+            )
+            if field_value is not None:
+                d[field_key] = field_value
+
+        return d
+    def __getattr__(self, item, *args):
+        self.__pydantic_extra__ = {item: {}}
+        print(args)
+        return {}
```
led02, line 80: `__getattr__` is called to return a dynamic attribute that is not defined by the class. Hence, when `config.get(...)` is evaluated, `get` is not found and `__getattr__` is called and expected to return some callable object.

We could implement this as `get(self, key, default=NO_DEFAULT)` like the following:

```suggestion
    _NO_DEFAULT = type("NoDefaultType", (str, ), {})

    def get(self, key, default=_NO_DEFAULT):
        try:
            return getattr(self, key)
        except AttributeError as e:
            if default is not self._NO_DEFAULT:
                return default
            raise e
```
Alternatively, we could forward the call to the dictionarized settings:

```suggestion
    def get(self, *args):
        data = self.model_dump()
        return data.get(*args)
```

However, I would prefer not to have any of this (see also https://github.com/hermes-hmc/hermes/pull/234#issuecomment-1878458975)

src/hermes/settings.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+    def __call__(self) -> Dict[str, Any]:
+        d: Dict[str, Any] = {}
+        for field_name, field in self.settings_cls.model_fields.items():
+            field_value, field_key, value_is_complex = self.get_field_value(
+                field, field_name
+            )
+            field_value = self.prepare_field_value(
+                field_name, field, field_value, value_is_complex
+            )
```
led02, line 72: These two statements could fit into one line if the maximum of 120 was asserted.

led02 (changes requested):
**note:** Not working yet, hence no approval.

I added some inline hints and also some general thoughts about the implementation.
I think the way forward here is to implement / adapt all the plugins that access configuration values (i.e., get rid of all the `get` calls).
Then we can in parallel further refactor the class-based plugin API (i.e., no central `settings.py` collecting all models anymore) and implement the CLI switch.

src/hermes/settings.py
```diff
@@ -0,0 +1,121 @@
...
+
+    model_config = SettingsConfigDict(env_file_encoding='utf-8', extra='allow')
+
+    harvest: HarvestSettings = HarvestSettings
+    deposit: DepositSettings = DepositSettings
+    config_path: ClassVar[pathlib.Path] = "hermes.toml" #TODO : Set config Path in cli
+
+    def __init__(self, conf_path):
+        super().__init__()
+        self.__class__.config_path = conf_path
```
led02, line 47: Forgot in my review: Why is this set on class-level? Are you sure this is what want? (Well, I can imagine why this might be a good idea, but I cannot read it in any documentation yet... ;) )

SKernchen:
Config is not in Context rigth now, maybe @sdruskat attempt for changing to argparse can help there

src/hermes/cli.py
```diff
@@ -86,8 +89,21 @@ def invoke(self, ctx: click.Context) -> t.Any:
 
         # Get the user provided working dir from the --path option or default to current working directory.
         working_path = ctx.params.get('path').absolute()
-
-        configure(ctx.params.get('config').absolute(), working_path)
+        config_path = ctx.params.get('config').absolute()
+        try:
+            with open(config_path, 'r'):
+                config = HermesSettings()
```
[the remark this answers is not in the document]
led02: **hint:** After reviewing and testing, this implementation does not behave as expected: While it checks for the availability of the `config_path`, it still reads the `hermes.toml`.

led02 (approved):

led02 (approved):

PR #235, opened 2023-12-11
Patch CFF author emails into CodeMeta data using cffconvert 2.0.0

sdruskat:
Until cffconvert is published with https://github.com/citation-file-format/cffconvert/pull/309, we'll lose email data for authors from CFF.

This PR provides a workaround by patching CodeMeta authors with their emails from the CFF data after creating the `Citation` object using cffconvert.


src/hermes/commands/harvest/cff.py
```diff
@@ -74,8 +74,19 @@ def _load_cff_from_file(cff_data: str) -> t.Any:
     return yaml.load(cff_data)
 
 
+def _patch_author_emails(cff_data, codemeta_str):
+    cff_authors = _load_cff_from_file(cff_data)["authors"]
```
poikilotherm, line 78: **nitpick**: I wonder if it would be more concise when not loading the data again by moving the call to this method from `_convert` up one level, behind calling `_convert` (line 66) and pass `cff_dict`.

src/hermes/commands/harvest/cff.py
```diff
@@ -74,8 +74,19 @@ def _load_cff_from_file(cff_data: str) -> t.Any:
...
+    codemeta = json.loads(codemeta_str)
+    for i, author in enumerate(cff_authors):
+        if "email" in author:
+            codemeta["author"][i]["email"] = author["email"]
+    return json.dumps(codemeta)
+
+
 def _convert_cff_to_codemeta(cff_data: str) -> t.Any:
     codemeta_str = Citation(cff_data).as_codemeta()
+    codemeta_str = _patch_author_emails(cff_data, codemeta_str)
```
poikilotherm, line 88: **nitpitck**: I would appreciate a comment this is a workaround, why it's here and a TODO to remove it later

(Also, see my comment above about potentially moving this patch method call up one level)

poikilotherm (approved):
Left two small comments which might be mostly bikeshedding. Otherwise LGTM!

led02 (changes requested):
I think the one comment by Oliver is very important (mark the hack as such / add a ToDo).

zyzzyxdonta (approved):

sdruskat:
> I think the one comment by Oliver is very important (mark the hack as such / add a ToDo).

@led02 Done. Please re-review.

led02:
Was ist denn da schon wieder mit Sphinx/RTD los?

led02 (approved):
What should I say... 🥇

PR #236, opened 2023-12-13
Release 0.9.0

sdruskat:
Prepares the 0.9.0 release to push to PyPI, following [the docs](https://docs.software-metadata.pub/en/latest/dev/contribute.html#stabilizing-the-codebase-and-making-releases).

## TODO

- [x] Fix Sphinx build
- [ ] Improve test coverage
- [ ] Decide whether we should require `Python >= 3.8` for this release (as discussed with somesy)


sdruskat:
@poikilotherm Gerne den Sphinx Build fixen.

sdruskat:
@zyzzyxdonta @led02 @poikilotherm @SKernchen Should we bump the Python requirement down to >= 3.8 at this stage already, as discussed with the somesy project?

led02:
Python 3.8 is not an option yet or did you already adapt the `match` statements?

zyzzyxdonta:
If our goal is to get this release done, I think Python 3.8 it can wait until hermes 1.0 😁 There are a number of things that need changing. E.g. type annotations using `list` and `dict`, certain string functions weren't available then, etc. We should take our time and set up the CI and static code analysis to find all problems rather than rushing it now.

PR #240, opened 2024-01-11
Fix Sphinx docs build

### LINKED ISSUE #239: Documentation builds fail with Auto API 2.0
We see in CI and in local builds, that `sphinx-autoapi` is somehow broken.

Some sources on the internet say upgrade the module, but this requires updating the whole Sphinx setup.

This issue is here to track the progress of that.

poikilotherm:
Closes #239


zyzzyxdonta (approved):
Works great! 👍🏻

Issue #244, opened 2024-03-06
Issues found during manual testing of 0.8.0 at deRSE24


poikilotherm:
- [x] No log output?
- [x] No complaints about invalid CFF file -> Logs are not enabled by default, missing init to ERR/WARN/INFO.
- [x] No complaints about non-existent or missing configuration options (example: harvest.cff.validate -> harvest.cff.enable_validation & harvest.from -> harvest.sources renamed) (Potential duplication of above)
- [x] #259
- [x] #260

poikilotherm:
Version given as `1.0` in CFF is interpreted as floating point and will give an error at deposit because not a valid string. We need to do a hard cast to string for the version

led02:
Logs and errors are now handled a bit saner... i.e., the errors are output to the log.error channel (which is by default printed on the console).

Still lot's of stuff to do that will be handled in different issues (i.e., #238).

sdruskat:
During the workshop, it was also mentioned that `hermes` should support older Python versions (i.e., 3.8).

I'm strictly against this, seeing that [EOLs are approaching fast](https://devguide.python.org/versions/).
3.8 will be unsupported from October 2024, 3.9 October 2025.

I'd argue that requiring 3.10 is actually feasible and shouldn't be changed.

led02:
I think it was also required to support SOMESY

sdruskat:
> I think 3.8 was also required to support SOMESY

True. I suggest asking somesy to update to 3.10 instead of supporting versions near the EOL. Or were there any (dependency-related) reasons for sticking with 3.8 (other than: a lot of people still use it)?

poikilotherm:
I think we can close this issue. Python 3.8 is EOL now and even 3.10 is going to be EOL in 1 year. See also https://endoflife.date/python

The logging stuff is tracked via #138, too.

The casting stuff is tracked via #281 / #291 .

PR #253, opened 2024-04-02
Fixes from the first release

led02:
The points are collected in #244.


led02:
Problems should be fixed, however at least the regressions from the issue should be covered by tests - ideally, even more!

@SKernchen @poikilotherm @zyzzyxdonta feel free to contribute further tests... ;)

(And also have a look at the `hermes_env` fixture that allows easily creation and setup of isolated hermes project envs.

sdruskat:
How did you fix #260?

src/hermes/commands/postprocess/invenio.py
SKernchen, line 1: Also put it in self.errors?

SKernchen (approved):

src/hermes/commands/postprocess/invenio.py
```diff
@@ -23,8 +23,8 @@ def config_record_id(ctx):
     try:
         conf.deposit.invenio.record_id = deposition['record_id']
         toml.dump(conf, open('hermes.toml', 'w'))
-    except KeyError:
-        raise RuntimeError("No deposit.invenio configuration available to store record id in") from None
+    except KeyError as e:
+        raise RuntimeError("No deposit.invenio configuration available to store record id in") from e
```
zyzzyxdonta, line 27: I wouldn't `raise from` the `KeyError` here. This will lead to the output

`exception was the direct cause of the following exception`

But it's not. The `KeyError` is just the symptom. The cause is the user not configuring the software right. We don't need to track where the exception came from. We already know.

src/hermes/commands/postprocess/invenio_rdm.py
```diff
@@ -22,5 +22,5 @@ def config_record_id(ctx):
     try:
         conf['deposit']['invenio_rdm']['record_id'] = deposition['record_id']
         toml.dump(conf, open('hermes.toml', 'w'))
-    except KeyError:
-        raise RuntimeError("No deposit.invenio configuration available to store record id in") from None
+    except KeyError as e:
+        raise RuntimeError("No deposit.invenio configuration available to store record id in") from e
```
zyzzyxdonta, line 26: Same as above.

src/hermes/logger.py
```diff
@@ -31,7 +31,7 @@
             'class': "logging.FileHandler",
             'formatter': "logfile",
             'level': "DEBUG",
-            'filename': "./.hermes/hermes.log",
+            'filename': "./hermes.log",
```
zyzzyxdonta, line 34: This file should be added to .gitignore

src/hermes/commands/cli.py
```diff
@@ -56,6 +58,28 @@ def main() -> None:
...
 
-    args.command.load_settings(args)
-    args.command.patch_settings(args)
-    args.command(args)
+    logger.init_logging()
+    log = logger.getLogger("hermes.cli")
+    log.debug("Running hermes with the following command line arguments: %s", args)
+
+    try:
+        log.info("Loading settings...")
```
zyzzyxdonta, line 66: I don't think there's value in showing this to the user. Of course settings are loaded. That's what every program does. 😅 

Maybe it would be useful to know where the settings are read from. But on the other hand, we can't even run the first stage without a hermes.toml. But in that case, the message is just debug and the user doesn't know if they don't look into the log file.

I would change this to `debug` in this particular case.

src/hermes/commands/postprocess/invenio_rdm.py
zyzzyxdonta, line 26:
```suggestion
        raise RuntimeError("No deposit.invenio_rdm configuration available to store record id in") from e
```

zyzzyxdonta (dismissed):

src/hermes/commands/cli.py
led02:
```suggestion
        log.debug("Loading settings...")
```

src/hermes/commands/postprocess/invenio.py
led02:
```suggestion
    except KeyError:
        raise RuntimeError("No deposit.invenio configuration available to store record id in")
```

Your such an incredible source of wisdom! Thanks for the hint

src/hermes/commands/postprocess/invenio_rdm.py
led02:
```suggestion
    except KeyError:
        raise RuntimeError("No deposit.invenio configuration available to store record id in")
```

src/hermes/logger.py
led02: Done

PR #273, opened 2024-09-03
Feature/init command

nheeb:
The init command is now working with Oauth. Zenodo Oauth saves the refresh token in the secret and deposit can now use a refresh token to get an access token.

### DIFF
--- a/docs/source/tutorials/automated-publication-with-ci.md
+++ b/docs/source/tutorials/automated-publication-with-ci.md
@@ -55,6 +55,8 @@ This also works with many Jupyter Hubs, no need for a Linux computer!
      hermes init
    ```
    Now simply follow the steps through the setup.
+7. When you're done, don't forget to add, commit and push the changes and new files you just made!
+
 
 ```{important}
 Be advised: during the setup, you can choose to grant access to all of your projects on your social coding platform.


src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,246 @@
...
+            else:
+                sc.echo("You better do it later or HERMES won't work properly.")
+                sc.echo("You can {} to create the file. Then move it into the project folder.".format(
+                    create_console_hyperlink(citation_cff_url, "click here")))
+        else:
+            sc.echo("Your project already contains a `CITATION.cff` file. Nice!")
+
+        # Creating the hermes.toml file
+        hermes_toml_raw_url = "https://raw.githubusercontent.com/nheeb/zenodo-test/main/hermes.toml"
+        download_file_from_url(hermes_toml_raw_url, os.path.join(os.getcwd(), "hermes.toml"))
```
zyzzyxdonta, line 167: This file should be part of this repo and then loaded via package data. There is no need to download it.

src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,246 @@
...
+            github_ci_template_raw_url = ("https://raw.githubusercontent.com/softwarepub/hermes/feature/init-command"
+                                          "/src/hermes/commands/init/tmp_hermes_github_to_zenodo.yml")
+            github_folder_path = os.path.join(os.getcwd(), ".github")
+            if not os.path.isdir(github_folder_path):
+                os.mkdir(github_folder_path)
+            workflows_folder_path = os.path.join(github_folder_path, "workflows")
+            if not os.path.isdir(workflows_folder_path):
+                os.mkdir(workflows_folder_path)
+            ci_file_path = os.path.join(workflows_folder_path, "hermes_github_to_zenodo.yml")
+            download_file_from_url(github_ci_template_raw_url, ci_file_path)
```
zyzzyxdonta, line 201: Same as above. No need to download.

src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,246 @@
...
+            # github_ci_template_raw_url = ("https://raw.githubusercontent.com/softwarepub/ci-templates/main"
+            #                               "/TEMPLATE_hermes_github_to_zenodo.yml")
+            github_ci_template_raw_url = ("https://raw.githubusercontent.com/softwarepub/hermes/feature/init-command"
+                                          "/src/hermes/commands/init/tmp_hermes_github_to_zenodo.yml")
+            github_folder_path = os.path.join(os.getcwd(), ".github")
+            if not os.path.isdir(github_folder_path):
+                os.mkdir(github_folder_path)
+            workflows_folder_path = os.path.join(github_folder_path, "workflows")
+            if not os.path.isdir(workflows_folder_path):
+                os.mkdir(workflows_folder_path)
```
zyzzyxdonta, line 199: This can be done nicely with pathlib. Something like:

```python
from pathlib import Path
(Path() / github_folder_path / "workflows").mkdir(parents=True, exist_ok=True)
```

src/hermes/commands/init/oauth_github.py
```diff
@@ -0,0 +1,73 @@
...
+
+    return "You have successfully logged in with GitHub! You can close this window now."
+
+
+def open_browser():
+    webbrowser.open(f"http://localhost:{local_port}")
+
+
+def run_flask():
+    serve(app, host='0.0.0.0', port=local_port)
```
zyzzyxdonta, line 47: `127.0.0.1` should be enough

src/hermes/commands/init/oauth_zenodo.py
```diff
@@ -0,0 +1,146 @@
...
+    # shutdown_func()
+    return "You have successfully logged in with Zenodo! You can close this window now."
+
+
+def open_browser():
+    webbrowser.open(f"http://localhost:{local_port}")
+
+
+def run_flask():
+    serve(app, host='0.0.0.0', port=local_port)
```
zyzzyxdonta, line 58: `127.0.0.1` should be enough

src/hermes/commands/init/oauth_github.py
```diff
@@ -0,0 +1,73 @@
+import os
+import webbrowser
+from flask import Flask, request, redirect
+from requests_oauthlib import OAuth2Session
+from threading import Timer, Thread, Event
+from waitress import serve
+
+local_port = 8333
+client_id = 'Ov23ctl0gNzr9smeVIHR'
+client_secret = 'd516303374f7e55189fe74fb2af77f31a965ad57'
```
zyzzyxdonta, line 10: Is this supposed to be here?

src/hermes/commands/init/oauth_zenodo.py
```diff
@@ -0,0 +1,146 @@
...
+import logging
+
+USING_SANDBOX = True
+
+local_port = 8334
+redirect_uri = 'http://localhost:' + str(local_port) + '/callback'
+sandbox_client_id = 'QJ8Q9GBI78uOdNmVNK1Vd0oAOJHqmYGvxRxiSFxt'
+sandbox_client_secret = 'nGuOqoDtd2tckP6lmQS3If3cY39lPLKLU8skcv72JeowNupMD2bnLparsGO9'
+real_client_id = 'L0d9HQVW4Ig9PnC6qh6zkOAwgvYy08GcmHJqVVvV'
+real_client_secret = '0HIvtC2D2aPvpq2W0GtfWdeivwkqvnvrOTGx14nUJA5lDXrEDSaQAnqxHbLH'
```
zyzzyxdonta, line 17: Is this supposed to be here?

pyproject.toml
```diff
@@ -50,6 +50,10 @@ pyparsing = "^3.0.9"
 requests = "^2.28.1"
 pydantic = "^2.5.1"
 pydantic-settings = "^2.1.0"
+flask = "^3.0.3"
+requests-oauthlib = "^2.0.0"
+waitress = "^3.0.0"
+pynacl = "^1.5.0"
```
zyzzyxdonta, line 56: I'm not a fan of pulling in flask and waitress. Receiving a request and responding with a redirect should be doable in Python's standard library.

zyzzyxdonta (changes requested):
I had a quick first look at this and wrote some comments. :-)

Some more that don't fit in the code:

- The Zenodo oauth application says "Application 'hermes-init-oauth' by '<your address here>' wants permission to access your 'd.pape@hzdr.de' account." It would be nice if this was the hermes mailbox address instead of your private one.
- The same for GitHub which currently says "hermes-init-oauth by [nheeb](https://github.com/nheeb)". Maybe the softwarepub group can be used here?
- During GitHub OAuth, I run into an error. Afterwards the CLI was stuck.

```
An error occurred during execution of init
ERROR:hermes.cli:An error occurred during execution of init
DEBUG:hermes.cli:Original exception was: Failed to retrieve public key: 404 {"message":"Not Found","documentation_url":"https://docs.github.com/rest/actions/secrets#get-a-repository-public-key","status":"404"}
```

Unfortunately, I'm clueless about OAuth, so I have no idea what is going wrong.

zyzzyxdonta:
Oh I see. I got a 404 because I gave a random git remote URI that doesn't exist but the code actually accesses my repo. 🤦🏻‍♂️

src/hermes/commands/init/oauth_github.py
nheeb: At least the secrets should be stored somewhere else in the end.

pyproject.toml
zyzzyxdonta: Maybe something like this?

```python
from http.server import BaseHTTPRequestHandler, HTTPServer

PORT = 5000
REDIRECT_URL = "https://www.hzdr.de"


class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        is_callback = self.path == "/callback"
        if is_callback:
            self.send_response(200)
            self.send_header("Content-Type", "text/plain")
        else:
            self.send_response(302)
            self.send_header("Location", REDIRECT_URL)
        self.end_headers()
        if is_callback:
            self.wfile.write(b"You can close this window now.")

    def log_request(self, code="-", size="-"):
        """Don't log anything."""
        pass


with HTTPServer(("127.0.0.1", PORT), Handler) as server:
    print(f"http://127.0.0.1:{PORT}", f"http://127.0.0.1:{PORT}/callback", sep="\n")
    server.serve_forever()
```

zyzzyxdonta: OK, the callback needs to be in `do_POST`, but the principle is the same.

zyzzyxdonta: OK, this should work for real this time 😄 

```python
from http.server import BaseHTTPRequestHandler, HTTPServer
from urllib.parse import parse_qs, urlparse

PORT = 5000
REDIRECT_URL = "https://www.hzdr.de"


class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        parsed = urlparse(self.path)
        is_callback = parsed.path == "/callback"

        if is_callback:
            self.send_response(200)
            self.send_header("Content-Type", "text/plain")
            print(parse_qs(parsed.query))
        else:
            self.send_response(302)
            self.send_header("Location", REDIRECT_URL)
        self.end_headers()
        if is_callback:
            self.wfile.write(b"You can close this window now.")

    def log_request(self, code="-", size="-"):
        pass


with HTTPServer(("127.0.0.1", PORT), Handler) as server:
    print(f"http://127.0.0.1:{PORT}", f"http://127.0.0.1:{PORT}/callback", sep="\n")
    server.serve_forever()
```

poikilotherm: I agree to avoid adding lots of additional deps if we can avoid it.

src/hermes/commands/init/__init__.py
```diff
@@ -0,0 +1,3 @@
+# SPDX-FileCopyrightText: 2022 German Aerospace Center (DLR)
```
poikilotherm, line 1: Please change this to an FZJ header.

poikilotherm (changes requested):
Overall this is a very good start. I agree with all of the points from @zyzzyxdonta and added some more :see_no_evil: 

@nheeb I also noticed flake8 is unhappy. You can execute it locally using `poetry run task flake8` with our configuration.

src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,244 @@
...
+        self.has_gitignore: bool = False
+        self.has_citation_cff = False
+        self.current_branch: str = ""
+
+
+def scout_current_folder() -> HermesInitFolderInfo:
+    info = HermesInitFolderInfo()
+    current_dir = os.getcwd()
+    info.absolute_path = str(current_dir)
+    info.has_git = os.path.isdir(os.path.join(current_dir, ".git"))
```
poikilotherm, line 39: Please also check if the Git command is actually available and fail gracefully - otherwise you'll run into trouble tryint to execute it.

src/hermes/commands/init/base.py
poikilotherm: I'm not so sure about this. Downloading it from our template repo enables us to update the template without making people update the hermes package. It also enables people to provide their own templates at some other URL.

The updated template is important within the context of Software CaRD: probably we will want to bury the old PR based workflow and replace it with sth noice.

src/hermes/commands/init/base.py
poikilotherm: Should it even be generated / serialized from a config object? (I agree we shouldn't download it)

src/hermes/commands/init/base.py
zyzzyxdonta: Why would the template need to change if hermes stays the same? What use would a custom template be without the functionality in hermes to fill in the blanks?

If we really want to download, it should be from a different repo. Otherwise we risk breaking the template when making changes to hermes.

src/hermes/commands/init/slim_click.py
```diff
@@ -0,0 +1,51 @@
...
+    for o in options:
+        char = o[0].lower()
+        description = o[1]
+        if char == default.lower():
+            description += " [default]"
+        description = o[1]
+        print(f"[{char}] {description}")
+    while True:
+        answer = input("Your choice: ").lower()
+        if answer in list(zip(*options))[0]:
```
zyzzyxdonta, line 46:
```suggestion
        if answer in dict(options).keys():
```

I think this is a bit easier to grasp.

src/hermes/commands/init/slim_click.py
```diff
@@ -0,0 +1,51 @@
...
+            return default
+        else:
+            print("Error: invalid input")
+
+
+def press_enter_to_continue(text: str = "Press ENTER to continue"):
+    input(text)
+
+
+def choose(text: str, options: list[tuple[str, str]], default: str = "") -> str:
```
zyzzyxdonta, line 34: Is there a specific reason you used list of tuples rather than `dict[str, str]`?

nheeb: You're right. Using a dict here is much better. :+1:

src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,244 @@
...
+
+from hermes.commands.base import HermesCommand
+import hermes.commands.init.oauth_github as oauth_github
+import hermes.commands.init.oauth_zenodo as oauth_zenodo
+import hermes.commands.init.github_secrets as github_secrets
+import hermes.commands.init.github_permissions as github_permissions
+import hermes.commands.init.slim_click as sc
+
+
+class HermesInitFolderInfo:
```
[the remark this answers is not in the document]
zyzzyxdonta: I don't have a strong opinion on where this class should be. However, I would make it a dataclass if it doesn't have any functionality.

src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,485 @@
...
+                info.current_branch = line.split()[1].strip()
+                break
+        if info.git_remote_url:
+            parsed_url = urlparse(info.git_remote_url)
+            info.git_base_url = f"{parsed_url.scheme}://{parsed_url.netloc}/"
+            sc.echo(f"git base url = {info.git_base_url}", debug=True)
+    if "github.com" in info.git_remote_url:
+        info.used_git_hoster = GitHoster.GitHub
+    elif "401" in subprocess.run(['curl', info.git_base_url + "api/v4/version"],
+                                 capture_output=True, text=True).stdout:
```
zyzzyxdonta, line 100: This request should be done in Python!

zyzzyxdonta (approved):
Overall, I think this looks pretty good. Just some small suggestions, nothing critical.

I'm still not happy with the oauth client details in the code because malicious actors might reuse them for their activities and they would show up as "hermes" when logging in. But if this is the only way to do it, then I guess there's no choice 🤷🏻‍♂️

src/hermes/commands/init/base.py
```diff
@@ -0,0 +1,480 @@
...
+                    "Please execute `hermes init` in the root directory of your git project.",
+                    formatting=sc.Formats.FAIL)
+            sys.exit()
+
+        # Abort if neither GitHub nor gitlab is used
+        if self.folder_info.used_git_hoster == GitHoster.Empty:
+            sc.echo("Your git project ({}) is not connected to github or gitlab. It is mandatory for HERMES to "
+                    "use one of those hosting services.".format(self.folder_info.git_remote_url),
+                    formatting=sc.Formats.FAIL)
+            sys.exit()
```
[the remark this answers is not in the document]
nheeb: Good question. On the one side we could of course give someone with that setup a hermes.toml.
On the other side that is already all we can do; leaving the setup process at ~80% unfinished. The CI/CD stuff but also the Deposit Platform related stuff requires Git.
I fear that it might give a false sense of completeness or encourage them to try setting up Hermes with a platform that might not even be compatible.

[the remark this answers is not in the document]
zyzzyxdonta: Why does the reposit stuff require GitLab or GitHub? 🤔 

Concerning the "completeness", we could print a message, something like:

>Skipped CI/CD setup (only available for GitLab/GitHub).

Maybe with a hint that they can write a feature request for their own Git host and/or CI platform.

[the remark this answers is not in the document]
nheeb: That sounds good. I will add it that way.

(For the Zenodo setup we need to save the access token as git project variable)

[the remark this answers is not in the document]
zyzzyxdonta: Oh, yup that's right. But if there's no CI/CD the token is not needed. We could just tell the user to create one themselves (or create it for them and print it to the terminal, but I'm not sure about this).

poikilotherm (approved):
Looks good to me! A good foundation to continue with! Good job @nheeb

PR #278, opened 2024-10-29
Feature/276 harvesting metadata from a provided repository url

Aidajafarbigloo:
Added the URL to the `hermes harvest` command. Now, the command `hermes harvest` harvest the metadata from the local repository, and `hermes harvest --url <URL>` allows harvesting metadata from the provided URL, with support for GitHub and GitLab repositories.

(e.g., `hermes harvest --url https://github.com/NFDI4Energy/SMECS`)


Aidajafarbigloo:
Harvesting metadata from the provided URL (GitHub/GitLab). Command: `hermes harvest --path <URL>`

Aidajafarbigloo:
@sferenz 
Could you please take a look at this pull request and share your feedback?

src/hermes/commands/base.py
```diff
@@ -131,7 +131,7 @@ def init_command_parser(self, command_parser: argparse.ArgumentParser) -> None:
     def load_settings(self, args: argparse.Namespace):
         """Load settings from the configuration file (passed in from command line)."""
 
-        toml_data = toml.load(args.path / args.config)
+        toml_data = toml.load("." / args.config)
```
sferenz, line 134: Does this still work if a regular path is given to HERMES?

src/hermes/commands/harvest/base.py
```diff
@@ -25,7 +25,7 @@ def __call__(self, command: HermesCommand) -> t.Tuple[t.Dict, t.Dict]:
         pass
 
 
-class HarvestSettings(BaseModel):
+class _HarvestSettings(BaseModel):
```
sferenz, line 28: Why did you rename this?

src/hermes/commands/harvest/util/remote_harvesting.py
```diff
@@ -0,0 +1,75 @@
...
+                            return _download_to_tempfile(file_url, filename)
+        else:
+            print(f"Unsupported repository URL: {repo_url}")
+            return None
+    except Exception as e:
+        print(f"Error fetching metadata from repository: {e}")
+        return None
+
+
+def _download_to_tempfile(url: str, filename: str) -> pathlib.Path:
```
sferenz, line 59: Do you delete the tempfiles later?

sferenz (changes requested):
Thanks for the nice code! Please have a look at the comments :)

src/hermes/commands/base.py
Aidajafarbigloo: Yes, specifying a directory path containing CFF or CodeMeta files is also acceptable. For example, the following command works:

`hermes harvest --path C:\path\to\your\directory`

src/hermes/commands/harvest/base.py
Aidajafarbigloo: I didn’t intend to change the class name. There was an issue with incorrectly pulling the original code for base.py from an updated version of it. This occurred due to a recent update in the settings classes, where all were made private in the develop branch of HERMES (commit a6c1a5e).

src/hermes/commands/harvest/util/remote_harvesting.py
Aidajafarbigloo: In the current code the temp files for CFF and CodeMeta are stored separately in `C:\Temp` on the local machine.

sferenz: Will these files be deleted after the extraction process?

sferenz: Yes, I think the temp files should be deleted at the end of the process.

Aidajafarbigloo: I updated the code to remove the temp files after the harvesting process. Could you please have a look on the changes?

src/hermes/commands/harvest/cff.py
```diff
@@ -45,6 +46,7 @@ def __call__(self, command: HermesHarvestCommand) -> t.Tuple[t.Dict, t.Dict]:
         # Read the content
         cff_data = cff_file.read_text()
 
+        remove_temp_file(cff_file)
```
sferenz, line 49: Is the CFF file also removed if it is not the temp directory?

src/hermes/commands/harvest/util/remote_harvesting.py
sferenz: Thanks!

sferenz (approved):

sferenz:
@sdruskat This pull request is ready to merge, can you please assign us a reviewer?

zyzzyxdonta (changes requested):
Thanks for your work!

I had a first look and I would like to suggest a slightly different approach. I think it would be beneficial to have the `--url` argument that you had (as indicated by your PR description). This would allow us to do the following:

1. Create a temporary directory
2. download the remote repository given by `--url` to this directory
3. overwrite `args.path` with the temporary directory path
4. run the normal harvesting step
5. delete the temporary directory

In this case there is no need to change anything in any of the plugins (I think). Only the base harvest command needs to worry about downloading and then deleting the files.

What do you think?

src/hermes/commands/harvest/util/remote_harvesting.py
```diff
@@ -0,0 +1,93 @@
...
+        with tempfile.NamedTemporaryFile(delete=False, suffix=f".{filename.split('.')[-1]}") as temp_file:
+            temp_file.write(content.encode("utf-8"))
+            print(f"Downloaded {filename} to {temp_file.name}")
+            return pathlib.Path(temp_file.name)
+    except Exception as e:
+        print(f"Error downloading {filename}: {e}")
+        return None
+
+
+def remove_temp_file(file_path: pathlib.Path, temp_dir: pathlib.Path = pathlib.Path("C:/Temp")):
```
zyzzyxdonta, line 85: `C:/Temp` is windows-specific. You could use `tempfile.TemporaryDirectory` and place the files in there. Then, instead of deleting the files one by one, you can use `.cleanup()` on the `TemporaryDirectory` object.

src/hermes/commands/harvest/util/remote_harvesting.py
```diff
@@ -0,0 +1,93 @@
...
+import typing as t
+import os
+
+def normalize_url(path: str) -> str:
+    """Normalize a given URL by correcting backslashes and fixing malformed HTTPS."""
+    corrected_url = path.replace("\\", "/")
+    return corrected_url.replace("https:/", "https://")
+
+
+def fetch_metadata_from_repo(repo_url: str, filename: str) -> t.Optional[pathlib.Path]:
```
zyzzyxdonta, line 21: This method makes multiple HTTP requests. I think it would be nice to use the hermes user agent, just to let the services know who we are. You can do something like:

```python
from hermes.utils import hermes_user_agent

session = requests.Session()
session.headers.update({"User-Agent": hermes_user_agent})
```

then use the session to make the requests:

```python
session.get(api_url)
```

zyzzyxdonta (changes requested):

src/hermes/commands/harvest/util/remote_harvesting.py
```diff
@@ -0,0 +1,157 @@
...
+                session.headers.update({"Authorization": f"token {token}"})
+            elif "gitlab" in repo_url:
+                session.headers.update({"PRIVATE-TOKEN": token})
+                
+        temp_dir_obj = tempfile.TemporaryDirectory()
+        temp_dir = pathlib.Path(temp_dir_obj.name)
+        
+        parsed_url = urlparse(repo_url)
+        
+        if "github.com" in repo_url:
```
zyzzyxdonta, line 49: This should test `if parsed_url.netloc == "github.com"`, otherwise some possible strange URLs like `https://gitlab.example.com/foo/github.com-migration-script` might be detected as GitHub.

src/hermes/commands/harvest/util/remote_harvesting.py
```diff
@@ -0,0 +1,157 @@
...
+            # GitHub API: List repository contents
+            api_url = repo_url.replace("github.com", "api.github.com/repos").rstrip("/") + "/contents"
+            response = session.get(api_url)
+            if response.status_code == 200:
+                for file_info in response.json():
+                    if file_info["name"] == filename:
+                        temp_file = _download_to_tempfile(file_info["download_url"], filename, temp_dir, session)
+                        return temp_file, temp_dir_obj
+        elif "gitlab" in parsed_url.netloc:
```
zyzzyxdonta, line 58: I think this will likely cover many cases. But we should have an additional list to check against. Something like `elif parsed_url.netloc in KNOWN_GITLAB_HOSTS or "gitlab" in parsed_url.netloc:` where `KNOWN_GITLAB_HOSTS = ["codebase.helmholtz.cloud", "jugit.fz-juelich.de", "git.ufz.de", "git.rwth-aachen.de"]` (just a quick example of instances I found in my browser history; there are certainly many more)

zyzzyxdonta (changes requested):
Some quick comments. Will have another look later or tomorrow. Thanks!

zyzzyxdonta:
Oh wait... Which is the correct pull request? This one or https://github.com/softwarepub/hermes/pull/473 ?

PR #288, opened 2024-12-09
Ha Ha Hack-A-Thon! 🐱‍💻

led02:


led02:
Tests are still failing but I prefer not to deactivate them (yet) but hope for someone (maybe me) to fix them.

After a chat with @sdruskat I would propose to introduce this change step by step:

1. Import the new data model (i.e., hermes.model.types, hermes.model.merge and hermes.model.context_manager). This is also a good opportunity to rename some of the stuff. ;)
2. Adapt the plugin implementations to use the new data model. This requires adapting the command classes and the plugins itself. However, the implementations from this branch can be used as a template. However, I would not advise to directly import the implementations because the provenance recording is still very sloppy and might need a straight re-engineering.
3. Get rid of legacy stuff.
4. Having the plugins ported, decisions need to be made:
  - The plugin class model is slightly improved in this branch. However, as the plugin model is meant to be refactored, this can be skipped.
  - A fairly well tampered (little over-engineered) implementation for recording provenance is available in hermes.model.prov. This can be fitted to the plugins spitting valuable provenance information.
  - Test should be adapted and  added .... massively!

Happy to discuss this with you.

SKernchen:
Done in #480

PR #291, opened 2024-12-09
Convert version to string in CFF

### LINKED ISSUE #281: Yaml changes version string to integer/float -> deposit error
At the HZDR Day we had a bug doing the deposit step. In the Cff the Version is given as an integer/float (from cff init). However in the deposit step the error occurs that 
`{"status": 400, "message": "A validation error occurred.", "errors": [{"field": "metadata.version", "messages": ["Not a valid string."]}]}
`.
It was solvable by putting the version in quotation marks.
Is that maybe changed in the newer cff version?
Should we make a workaround by changing the type in the harvest step?

SKernchen:
Closes #281

### DIFF
--- a/src/hermes/commands/harvest/cff.py
+++ b/src/hermes/commands/harvest/cff.py
@@ -47,6 +47,7 @@ class CffHarvestPlugin(HermesHarvestPlugin):
 
         # Validate the content to be correct CFF
         cff_dict = self._load_cff_from_file(cff_data)
+
         if command.settings.cff.enable_validation and not self._validate(cff_file, cff_dict):
             raise HermesValidationError(cff_file)
 
@@ -54,6 +55,7 @@ class CffHarvestPlugin(HermesHarvestPlugin):
         codemeta_dict = self._convert_cff_to_codemeta(cff_data)
         # TODO Replace the following temp patch for #112 once there is a new cffconvert version with cffconvert#309
         codemeta_dict = self._patch_author_emails(cff_dict, codemeta_dict)
+        codemeta_dict["version"] = str(codemeta_dict["version"])    # Convert Version to string
 
         return codemeta_dict, {'local_path': str(cff_file)}
 


src/hermes/commands/harvest/cff.py
```diff
@@ -47,13 +47,15 @@ def __call__(self, command: HermesHarvestCommand) -> t.Tuple[t.Dict, t.Dict]:
...
         cff_dict = self._load_cff_from_file(cff_data)
+
         if command.settings.cff.enable_validation and not self._validate(cff_file, cff_dict):
             raise HermesValidationError(cff_file)
 
         # Convert to CodeMeta using cffconvert
         codemeta_dict = self._convert_cff_to_codemeta(cff_data)
         # TODO Replace the following temp patch for #112 once there is a new cffconvert version with cffconvert#309
         codemeta_dict = self._patch_author_emails(cff_dict, codemeta_dict)
+        codemeta_dict["version"] = str(codemeta_dict["version"])    # Convert Version to string
```
zyzzyxdonta, line 58: Do we know for certain that the `version` field exists? 🤔 I would guard this with an `if "version" in codmeta_dict`

sdruskat: > Do we know for certain that the `version` field exists? 🤔 I would guard this with an `if "version" in codmeta_dict`

No, we don't know that. It's not a required field in CFF.

SKernchen: Should we leave a comment (maybe in a logger or so) that the version is required for some deposition platforms?

zyzzyxdonta: Is it? On Invenio(-RDM) it defaults to 1.0

zyzzyxdonta (approved):

src/hermes/commands/harvest/cff.py
SKernchen: ah ok then not^^

PR #299, opened 2024-12-11
Add "Plugin marketplace" infrastructure to hermes docs and codebase

### LINKED ISSUE #292: Plugin Marketplace
- Plugin Marketplace:
    - Wie sollen neue Plugins bereitgestellt und sichtbar gemacht werden (von unserer Seite aus)?
        - Optionen z.B.: Marketplace-Übersicht, Namespaces (PyPI)
    - Sollen Plugins für verschiedene Steps verschieden behandelt werden? Z.B.:
        - Harvesting komplett in core integrieren?
    - `init`-Angebote
    - Metadatensatz für HERMES Plugins
- **Decision:**
        1. Liste (kuratiert)
        2. Metadatanformat + Issue-Template

zyzzyxdonta:
This pull requests adds a list of available plugins to the hermes docs page that can act as a first basic "marketplace".

TODO:

- [x] Add data extraction script
- [x] Add issue template for people who want to contribute their own plugin to the list
- [ ] List of plugins complete?
- [x] How to signify "builtin" in schema.org markup? → `schema:isPartOf`
- [x] How to add the repo URL to the markup? It seems only `SoftwareSourceCode targetProduct SoftwareApplication .` is possible. So we would need a `{"@type": "SoftwareSourceCode", "targetProduct": {"@type": "SoftwareApplication"}}`. This would be valid but not as nice as having all the plugins as top-level elements. Alternative: `schema:url`
- [ ] ~~Add markup for `<hermes> softwareAddOn <each plugin> .` ?~~ → Doesn't work well with "one script tag per plugin" paradigm.

The first simple design looks like this:

Screenshot_20241211_092557

In addition to the human-readable text, machine-readable markup is added. Schema.org markup was used. The markup on the page can be checked using the browser extension "OpenLink Structured Data Sniffer" or by pasting the source code of the compiled page into https://validator.schema.org/

Closes #292

### DIFF
--- a/src/hermes/commands/marketplace.py
+++ b/src/hermes/commands/marketplace.py
@@ -118,11 +118,10 @@ def main():
         row_sep = "-" * (17 + max_name_len + max_loc_len)
         print("HERMES step   " + "Plugin name" + (" " * (max_name_len - 8)) + "Plugin location")
         print(row_sep)
-        name_alignment = max(map(lambda plugin: len(plugin.name), parser.plugins))
         plugins_sorted = _sort_plugins_by_step(parser.plugins)
         for step in plugins_sorted.keys():
             for plugin in plugins_sorted[step]:
-                print(f"{step:>11}   {plugin.name:{name_alignment}}   {_plugin_loc(plugin)}")
+                print(f"{step:>11}   {plugin.name:{max_name_len}}   {_plugin_loc(plugin)}")
         print(row_sep)
         print()
 


sdruskat:
@zyzzyxdonta Nice one! 🚀 

I like this simple impl! A few comments:

- One thing I'd like to see added to the metadata is the **type of the plugin**, i.e., the step that this plugin targets. This will help filtering and could be used to display plugins of the same type under dedicated subheadings.
- In a next step, perhaps we could add another optional **field that describes the filetypes** (via, e.g., filename or schema) that a plugin works on? This makes sense mainly for harvest plugins, but I'd rather have an optional field that goes unused than introduce plugin metadata inheritance ;).
- At some point, we should probably provide a **JSON schema** for the plugin metadata?
- For now, having a centralized list of plugins makes sense. If at some point this becomes feasible, it could be better to let each plugin have its own metadata in its repo, where the marketplace retrieves it from. This way, plugin authors keep the metadata where the plugin is developed for easier updates. *And* we have less maintenance work to do, and a simple automated registration of a new plugin would suffice (register via issue, closing issue triggers CI to pull metadata and re-render marketplace).

Thoughts?

zyzzyxdonta:
> * One thing I'd like to see added to the metadata is the **type of the plugin**, i.e., the step that this plugin targets. This will help filtering and could be used to display plugins of the same type under dedicated subheadings.

Done.

Screenshot_20241211_120537

zyzzyxdonta:
> I think it's important to have the steps annotated but I think this is not a one-to-one relationship.

I've taken this into account by allowing multiple steps to be listed. If a plugin targets multiple steps, it appears in all of the related lists for now.

zyzzyxdonta:
I wrote a simple Sphinx extension to get rid of the horrible templating of the JSON-LD markup 🙈 

Builtin extensions are now marked using `schema:isPartOf <hermes>`.

zyzzyxdonta:
I feel like we could have a common data model (e.g. pydantic model) between the Sphinx extension and the `hermes-marketplace` command 👀

zyzzyxdonta:
> * At some point, we should probably provide a **JSON schema** for the plugin metadata?

Done

zyzzyxdonta:
This is what the issue template looks like (though some fields will be shown as required which doesn't work in the private repo I used for testing):

Screenshot_20241217_111342

zyzzyxdonta:
One problem: If we ship this to users, the URL `https://hermes.software-metadata.pub#plugins` needs to exist forever and we can't move the marketplace elsewhere. Should we create a redirect in read the docs, e.g. for `/marketplace` (that redirects to `/#plugins` for now) and then use `https://hermes.software-metadata.pub/marketplace` instead?

zyzzyxdonta:
> One problem: If we ship this to users, the URL `https://hermes.software-metadata.pub#plugins` needs to exist forever and we can't move the marketplace elsewhere. Should we create a redirect in read the docs, e.g. for `/marketplace` (that redirects to `/#plugins` for now) and then use `https://hermes.software-metadata.pub/marketplace` in the source code instead? I think this should be possible with ["Exact redirects"](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#exact-redirects)

Done

sdruskat (approved):
Thanks @zyzzyxdonta, I'm really loving this setup 😍.

As discussed, I've pushed one more commit improving the output of the `hermes-marketplace` command slightly by including step information and sorting plugins under each step they target.

Happy for you to merge this if you approve of my changes.

A screenshot of the updated output:

grafik

PR #314, opened 2025-02-26
Add version subcommand

### LINKED ISSUE #280: hermes --version
Now that we will provide multiple hermes versions, we should support the --version comand.
That would also be good for testing if the installation was correct.

poikilotherm:
Closes #280


nheeb (changes requested):
hermes.toml should be restored. The rest works and it's already visible in hermes help. nice

PR #321, opened 2025-02-28
Update pyproject.toml

nheeb:


zyzzyxdonta:
Should we use a Poetry plugin to automate this? https://github.com/mtkennerly/poetry-dynamic-versioning

nheeb:
> Should we use a Poetry plugin to automate this? https://github.com/mtkennerly/poetry-dynamic-versioning

Sounds like a good idea

led02:
I have a strong opinion against that!

- We need to track the correct version number in our CITATION.cff anyways
- With the traditional way of updating the version numbers in release and development branches it is clear to see towards what next release we are developing (i.e., is it a major update?)

What I still se useful is an automation that, e.g., "replicates" the version number from a single source of truth (CITATION.cff) and automatically creates branches or tags from it.

poikilotherm (approved):
Solid!

Issue #308, opened 2025-02-03
`hermes init` fails with cryptic error message in non-git directory


zyzzyxdonta:
When running `hermes init` in a non-git directory, it fails with:

```
No connection adapters were found for ':///api/v4/version'
```

This is due to `urlparse()` not "failing" but rather returning empty strings, here: https://github.com/softwarepub/hermes/blob/a67e2f95c1216236d7ca540f14558ceae67b7bed/src/hermes/commands/init/base.py#L96

It might make sense to wrap `urlparse()` and raise an exception if either of `scheme`  or `netloc` is not existent. (Or abort searching for git stuff way earlier.)

poikilotherm:
Closes by now merged #309

dnlbauer:
It also happens on an existing project from me that has multiple remotes configures in the git config:

```bash
git clone https://github.com/dnlbauer/django-signposting
cd django-signplosting
hermes init # this works!

# add second remote
git remote add secondary http://example.com/django-signposting
hermes init # fails
```

poikilotherm:
Thanks for reporting back! I'm assigning @nheeb to look into this 😄

PR #323, opened 2025-02-28
Feature/308 init with remotes

nheeb:
The new #308 problems should be fixed.

### DIFF
--- a/src/hermes/commands/init/base.py
+++ b/src/hermes/commands/init/base.py
@@ -56,6 +56,7 @@ class HermesInitFolderInfo:
     def __init__(self):
         self.absolute_path: str = ""
         self.has_git_folder: bool = False
+        self.has_multiple_remotes: bool = False
         self.git_remote_url: str = ""
         self.git_base_url: str = ""
         self.used_git_hoster: GitHoster = GitHoster.Empty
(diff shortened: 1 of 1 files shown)


SKernchen (dismissed):
When I tried to simulate this code the setting 
`Allow GitHub Actions to create and approve pull requests`
was not tickeled, so the pull request failed.
But the rest worked fine so I will approve the PR and create an issue.

src/hermes/commands/init/util/git_info.py
```diff
@@ -0,0 +1,106 @@
...
+    :return: The command's output.
+    """
+    # Validate or get default cwd
+    cwd = get_valid_cwd(cwd)
+    # Get command list from string
+    command_list = [s for s in command.split(" ") if s != ""]
+    if command_list[0] != "git":
+        command_list.insert(0, "git")
+    # Run subprocess
+    result = subprocess.run(command_list, cwd=cwd, capture_output=True, text=True)
```
zyzzyxdonta, line 48: I think it would be a good idea to have a common interface for git commands. Or at least to collect them all in some kind of util file. Because calling git is not trivial; see [here](https://github.com/softwarepub/hermes/pull/217).

But I think we can do this later and keep this here for now.

src/hermes/commands/init/base.py
```diff
@@ -53,75 +52,37 @@ class DepositPlatform(Enum):
...
+        # self.git_remote_url: str = ""
+        # self.git_base_url: str = ""
+        # self.used_git_hoster: GitHoster = GitHoster.Empty
         self.has_hermes_toml: bool = False
         self.has_gitignore: bool = False
         self.has_citation_cff: bool = False
         self.has_readme: bool = False
-        self.current_branch: str = ""
+        # self.current_branch: str = ""
         self.current_dir: str = ""
```
zyzzyxdonta, line 70: Commented-out code should be removed :-)

zyzzyxdonta (approved):
This makes sense to me and all of the setups I tested worked 👍🏻

PR #327, opened 2025-03-12
Fix broken HERMES DOI in plugin marketplace

### LINKED ISSUE #326: Broken HERMES DOI in marketplace plugin metadata
Builtin HERMES plugins are identified in the [marketplace](https://hermes.software-metadata.pub/marketplace) metadata by their `isPartOf` attribute which points to HERMES' DOI. However, this DOI is not rendered as an absolute URL (`https://doi.org/10.5281/zenodo.13311079`), but as a relative one (`10.5281/zenodo.13311079`). This is incorrect as it would be interpreted as belonging to the documentation domain (`https://hermes.software-metadata.pub/en/latest/10.5281/zenodo.13311079`). Thus, the absolute URL for the DOI should be rendered into the page.

zyzzyxdonta:
This pull request fixes the broken HERMES DOI used all over the plugin marketplace (#326).

Additionally, it adds a more reliable method of detecting builtin plugins.

BREAKING CHANGE: By changing the JSON-LD metadata in the docs page, this change breaks the `hermes-marketplace` command in HERMES v0.8.1 and v0.9.0. The command will continue to run successfully, but it will not be able to detect "builtin" plugins any more, and simply show them without a location.

Closes #326.

### DIFF
--- a/src/hermes/commands/marketplace.py
+++ b/src/hermes/commands/marketplace.py
@@ -4,6 +4,8 @@
 # SPDX-FileContributor: Stephan Druskat
 
 """Basic CLI to list plugins from the Hermes marketplace."""
+
+from functools import cache
 from html.parser import HTMLParser
 from typing import List, Optional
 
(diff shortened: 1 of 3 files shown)


zyzzyxdonta:
@poikilotherm tests were added.

nheeb (approved):
The changes are understandable and valid :+1:

Issue #298, opened 2024-12-09
init: create and implement hooks to make plugins create their config and ask questions during setup


poikilotherm:
Related to #292

PR #331, opened 2025-04-28
Feature/298 init install plugins

### LINKED ISSUE #328: hermes init cancelation should remove created files
If the user cancels initialization, then the .hermes repository is still created and in the directory.
Maybe after cancelation a hermes clean can delete that.

Same for the hermes.log, but there the clean command would not help.

nheeb:
Plugins can be installed with init
Closes #328 
Related to #298 (plugins don't really want to ask questions yet)

### DIFF
--- a/src/hermes/commands/init/base.py
+++ b/src/hermes/commands/init/base.py
@@ -19,6 +19,7 @@ from pydantic import BaseModel
 from requests import HTTPError
 
 import hermes.commands.init.util.slim_click as sc
+from hermes.commands import marketplace
 from hermes.commands.base import HermesCommand, HermesPlugin
 from hermes.commands.init.util import (connect_github, connect_gitlab,
                                        connect_zenodo, git_info)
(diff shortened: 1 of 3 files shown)


zyzzyxdonta (changes requested):

src/hermes/commands/init/base.py
```diff
@@ -173,8 +181,21 @@ def __init__(self, parser: argparse.ArgumentParser):
...
+                "sources": ["cff"]
+            },
+            "deposit": {
+                "target": "invenio_rdm",
+                "invenio_rdm": {
+                    "site_url": "",
+                    "access_right": "open"
+                 }
+            }
+        }
```
zyzzyxdonta, line 195: This feels like we should use `hermes.commands.base._HermesSettings` instead. But I don't see how 😅

zyzzyxdonta (approved):
Makes sense to me 👍🏻

PR #349, opened 2025-07-01
Allow initial publications on Rodare

### LINKED ISSUE #348: Make `--initial` deposits work on Rodare
Currently, the Rodare deposit plugin prohibits making initial deposits. Instead, we should clearly communicate that a registration in Robis is required first. This should be shown to users when they attempt to make an initial deposit but no Robis ID is given in the config

zyzzyxdonta:
Closes #348


nheeb (approved):
Nothing to complain :+1:

Issue #373, opened 2025-07-10
Error in error message in ValueError


SKernchen:
In
https://github.com/softwarepub/hermes/blob/feature/153-refactor-datamodel/src/hermes/model/context_manager.py#L60
the ValueError won't raise because the list index is out of range when you pop the last element.
And the value right now is wrong because the last step was already popped and so it shows the new last step, therefore the step before.

SKernchen:
Also the error message can differentiate from the error.
It can be a normal ValueError with a false key and still be displayed that the step order is the problem

PR #382, opened 2025-07-11
Tests for HermesContext and HermesCache

SKernchen:


led02 (changes requested):
Nice and well collected test cases, also very pythonic!

See inline comments for minor changes.

test/hermes_test/__init__.py
```diff
@@ -1,3 +0,0 @@
-# SPDX-FileCopyrightText: 2022 German Aerospace Center (DLR)
-#
-# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
SKernchen: Yes we can keep it, I think it changed automatically, because of recent open files.

src/hermes/model/context_manager.py
```diff
@@ -54,11 +54,21 @@ def __init__(self, project_dir: pathlib.Path = pathlib.Path.cwd()):
     def prepare_step(self, step: str, *depends: str) -> None:
         self._current_step.append(step)
 
+    ''' @FIXME #373:
```
sdruskat, line 57:
```suggestion
    '''
    FIXME: Implement this to fix #373
```

@SKernchen re your question: This is how I imagined we mark things we come across that need to be fixed during the next refactoring cycle, but not right now. 🤷

In the end, the style doesn't matter, as long as it's somewhat consistent. I usually use plain `FIXME` which gets highlighted in all IDEs I use, but we can use `@FIXME` if you prefer...

src/hermes/model/context_manager.py
```diff
@@ -54,11 +54,22 @@ def __init__(self, project_dir: pathlib.Path = pathlib.Path.cwd()):
...
         self._current_step.append(step)
 
+    '''
+    FIXME: Implement this to fix #373
+
+    def finalize_step(self, step: str) -> None:
+        current_step = self._current_step[-1]
+        if current_step != step:
+            raise ValueError(f"Cannot end step {step} while in {self._current_step[-1]}.")
+        self._current_step.pop()
```
led02, line 64: Why do you put the "fix" into the comment instead of putting it into the implementation?

The however, the first line will still throw an IndexError for empty lists.

src/hermes/model/context_manager.py
```diff
@@ -54,11 +54,22 @@ def __init__(self, project_dir: pathlib.Path = pathlib.Path.cwd()):
...
+    '''
     def finalize_step(self, step: str) -> None:
         current_step = self._current_step.pop()
         if current_step != step:
             raise ValueError("Cannot end step %s while in %s.", step, self._current_step[-1])
 
+
+
```
led02, line 72: PEP8-Violation 🚨

led02 (approved):
Minor fixes which can be addressed in other referenced issues.

src/hermes/model/context_manager.py
sdruskat: Documented in #373.

test/hermes_test/__init__.py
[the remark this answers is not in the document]
sdruskat: The [FAQs](https://reuse.software/faq/#uncopyrightable) suggest that it's fine to use either the default copyright, or `CC0-1.0` for license and `NONE` for copyright holder. So I think that keeping as is is fine.

src/hermes/model/context_manager.py
```diff
@@ -55,10 +55,19 @@ def prepare_step(self, step: str, *depends: str) -> None:
         self._current_step.append(step)
 
     def finalize_step(self, step: str) -> None:
-        current_step = self._current_step.pop()
+        if len(self._current_step) < 1:
+            raise ValueError("There is no step to end.")
+        current_step = self._current_step[-1]
         if current_step != step:
```
led02, line 61:
```suggestion
        if self._current_step[-1] != step:
```

No need to keep the local variable... or use it in the f-string in 62, too.

led02 (approved):

PR #398, opened 2025-07-24
Refactor/381 test ld container

### LINKED ISSUE #381: Test `ld_container`
Test all functions comprehensively.

SKernchen:
Test ld_container


led02:
I see some issues here:
- #400 Reflect the abstractness of `ld_container` in the implementation
- #401 Decide on the future of `path`
- #402 Documentation needs to be ... "improved"

test/hermes_test/model/types/test_ld_container.py
```diff
@@ -0,0 +1,88 @@
...
+
+        cont = ld_container({"spam": [{"@value": "bacon"}]})
+        cont.add_context([self.url])
+
+        assert cont.context == [self.url]
+        assert cont.full_context == [self.url]
+
+    def test_container_parent(self):
+        cont_parent = ld_container({"ham": [{"@value": "eggs"}]})
+        cont = ld_container({"spam": [{"@value": "bacon"}]}, parent=cont_parent)
```
led02, line 57: Construction principle for parent/child containers should be the following (even though, you should not construct `ld_container` instances due to their abstractness and internal nature ;) )

```python
child_data = [{"spam": [{"@value": "bacon"}]}]
parent_data = [{"ham": child_data}]

cont_parent = ld_container(parent_data)
cont = ld_container(child_data, parent=cont_parent, key="ham")
```

Important information that is hidden here (and should be explicited in documentation aritfacts):
- `ld_container` classes work as proxies and reflect the content and structure of the maintained datasets.
- All parent-child-relations of `ld_containers` need to be existent in the base dataset for this to work. (The opposite is not required).
- All datasets that are maintained by the `ld_container` are handled by reference! I.e., if you change something in the ld_container, it will be changed in the dataset directly. There is no place where a container creates a copy of the base dataset.
- If a container has a `parent`, it is expected to have either `index` or `key` (but not both).

SKernchen: That helps a lot, thank you!

test/hermes_test/model/types/test_ld_container.py
```diff
@@ -67,22 +69,38 @@ def test_container_full_context_and_path(self, httpserver):
...
+        cont = ld_container([{"spam": [{"@value": "bacon", "@id": "ham", "@type": ["@id"]}]}])
+        assert cont._to_python("spam", [{"@value": "bacon"}]) == 'bacon'
+        assert cont._to_python("@id", "ham") == "ham"
+
+        cont.active_ctx['_uuid'] = str(uuid.uuid1())    # FIXME: 406
+        assert cont._to_python("@type", ["@id"]) == '@id'
+
+    def test_to_expanded(self):
+        cont = ld_container([{"spam": [{"@value": "bacon", "@id": "ham", "@type": "@id"}]}])
+        cont.active_ctx['_uuid'] = str(uuid.uuid1())    # FIXME: 406
```
led02, line 103: Okay, `active_ctx` is something internal to `pyld` and is used here ... as some kind of shortcut. Problem is, that is even for `pyld` something internal and hence not (very well) documented.

I still see that #406 is an issue, though. Again, a stacktrace would be helpful (but I can also run the tests myself for further analysis).

The other question is, how `_to_expanded_json` should behave on undefined properties...

PR #421, opened 2025-08-07
Extend #403: Mark xfailing tests, raise errors

sdruskat:
This PR 

- implements error raising based on the assumptions in https://github.com/softwarepub/hermes/pull/403#issue-3268247506 (i.e., that `_getitem__` raises on bad input;
- marks those tests as expected to fail that test unimplemented behaviour that is assumed to be correct.

PR also merges #420 prematurely to be able to work with the new exception.


SKernchen (approved):
Still have to talk about refactoring stuff

Issue #363, opened 2025-07-09
Record provenance in a standardized way


sdruskat:
@led02 [suggested](https://github.com/softwarepub/hermes/issues/153#issuecomment-2975712552) using W3 PROV-O for this, which seems like a good idea.

There is an initial basic implementation in #288, but this needs more conceptual work.

This issue tracks this.

- Related: #153, #336 

### TODOs

- [x] Check with PROV pros for missing processes & correct implementation (@sdruskat)
- [ ] Add curation activity
- [x] Fix typo (duplicate curation data)
- [x] Find solution for recording provenance from deposit substeps
- [x] Design post-process command

- [x] Where are settings recorded? (probably in the plugin agents)

Issue #393, opened 2025-07-17
Develop and publish `hermes` and related vocabularies


sdruskat:
- Related: #379
 
### TODOs

- [ ] Decide if and how to separate vocabularies with different concerns ("content", "prov", etc., see current implementation details in #288).
- [ ] Decide how to develop, version and publish vocabularies
    - E.g., if we extract all plugins, should they all define own vocabularies (see, e.g., #405 for `pyproject.toml`-specific metadata)? Or, under which circumstances can they contribute to hermes' vocabs? (Example: terms defined by the git plugin: gitAuthor, etc.; PyPI classifiers may be mapped to different terms (see, #405, e.g., `Development Status` can be mapped to `developmentStatus`, etc.)
    - Decide on IRI scheme that is ideally backed by a long-living URL (e.g., do we agree on `https://schema.software-metadata.pub/<schema-name>/<schema-version>/<term>`?) See also #392 for, e.g., use of "schema".

For publicizing the schema(s) and making it persistently accessible, I'm preparing a W3ID here: https://github.com/perma-id/w3id.org/pull/5152.

PR #442, opened 2025-10-17
Document standardized provenance recording

notactuallyfinn:
This PR adds an adr that documents the current state of the concept for recording provenance information.
It also contains a diagram that visualizes the graphic of the information that will be recorded in the future according to the current state of discussion.
All contents of this PR are only up to discussion.
See also #363.

### DIFF
--- a/docs/adr/hermes-prov-diagram/hermes-prov.drawio.license
+++ b/docs/adr/hermes-prov-diagram/hermes-prov.drawio.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: 2025 German Aerospace Center (DLR)
+
+SPDX-License-Identifier: CC-BY-SA-4.0
--- a/docs/adr/hermes-prov-diagram/hermes-prov.svg.license
+++ b/docs/adr/hermes-prov-diagram/hermes-prov.svg.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: 2025 German Aerospace Center (DLR)
+
+SPDX-License-Identifier: CC-BY-SA-4.0


zyzzyxdonta:
Another idea for visualization: Maybe you could have data entities and activities running "in parallel" rather than alternating in a single line. Example without arrows:

```
--------------
|            |
|CITATION.cff|
|            |
--------------
                        --------------
                        |            |
                        |    load    |
                        |            |
                        --------------
--------------
|            |
|CITATION Py |
|            |
--------------
                        --------------
                        |            |
                        |    map     |
                        |            |
                        --------------
--------------
|            |
| software   |
| metadata   |
--------------
                        --------------
                        |            |
                        |   write    |
                        |            |
                        --------------
```

sdruskat:
> We/ I have considert `HermesCache` to be an `Agent` (more specifically a `SoftwareAgent`) because it bears the responsibility to store a SoftwareMetadata object and load the file again.

I agree. `HermesCache` here is the concrete instance of the class that has the `write` method, not the abstract idea of "we have a cache where stuff goes." Hence, I think this is a misunderstanding, as the [statement](https://github.com/softwarepub/hermes/pull/442#pullrequestreview-3376086234) "the fact that codemeta.json is stored inside the HermesCache" is false (`HermesCache` is the object at runtime, not the drive directory).

> I think that `codemeta.json` `wasAttributedTo` `HermesCache` because `HermesCache` stores the SoftwareMetadata object in `codemeta.json` and the only difference between those two `Entities` (`codemeta.json` and the SoftwareMetadata object) is that on is stored in a file and the other is not. And who else but the `HermesCache` could be associated with the storing/ loading?

I also agree here.

> All the provenance data will probably be stored as JSON-LD therefor I guess that all other properties could be modelled with the schema and codemeta vocab.

Yes, it will be stored as JSON-LD, using the [PROV-JSONLD Serialization](https://www.w3.org/submissions/2024/SUBM-prov-jsonld-20240825/) ([schema](https://www.w3.org/submissions/2024/SUBM-prov-jsonld-20240825/schema.json) | [context](https://www.w3.org/submissions/2024/SUBM-prov-jsonld-20240825/context.jsonld)). This should probably be in the ADR.

sdruskat:
> Another idea for visualization: Maybe you could have data entities and activities running "in parallel" rather than alternating in a single line.

This is an example of how I think the swimlanes could look like (horizontal is perhaps not ideal though).

swimlanes

It'll be a looooong diagram, but perhaps it's worth it? 🤷 
For user docs, we should use only snippets of this I guess (around the size of the example).

Here's a [drawio XML file](https://github.com/user-attachments/files/23418580/pro-o-swimlanes.drawio).

zyzzyxdonta:
> `HermesCache` here is the concrete instance of the class that has the `write` method, not the abstract idea of "we have a cache where stuff goes." Hence, I think this is a misunderstanding, as the [statement](https://github.com/softwarepub/hermes/pull/442#pullrequestreview-3376086234) "the fact that codemeta.json is stored inside the HermesCache" is false (`HermesCache` is the object at runtime, not the drive directory).

Yes, this was not clear to me. There is currently no such thing as `HermesCache` so I didn't recognize it as a class/object name. (Though the PascalCase could have given it away.) I only considered hermes, the plugins, and people as the agents. Having ~~a class~~ an object as an agent is way more detailed than I would have done this 😄

sdruskat:
> Sure. But neither models the concept of functions and parameters. The closest thing would be [`Action`](https://schema.org/Action). Are you planning to use [`object`](https://schema.org/object) for the function arguments?
> 
> (To model functions, args, kwargs, I mean)

That's a good point, but I don't think we're yet clear on the actual schema vocabularies to use. FWIW, we have the same issue elsewhere, so perhaps this needs to be solved as part of #393.

docs/adr/0014-standardized-provenance-recording.md
```diff
@@ -0,0 +1,62 @@
...
+* Date: 2025-10-17
+
+Technical story: 
+* https://github.com/softwarepub/hermes/pull/442
+* https://github.com/softwarepub/hermes/issues/363
+
+## Context and Problem Statement
+
+To consolidate traceability of the metadata, and resolution based on metadata sources in case of duplicates, etc., we need to record the provenance of metadata values in a __standardized__ way.
+Additionally we use the [PROV-O ontology](https://www.w3.org/TR/prov-o/) and [JSON-LD](https://www.w3.org/TR/json-ld/) and want that HERMES records as much of the provenance as possible to not overcomplicate plugin development.
```
sdruskat:
```suggestion
To achieve this, we use the [PROV-O ontology](https://www.w3.org/TR/prov-o/) serialized as [JSON-LD](https://www.w3.org/TR/json-ld/). Additionally, HERMES should make it possible to record as much of the provenance as possible *centrally*, i.e., as part of the core codebase. This is to keep plugin developers from having to supply their own provenance solutions.
```

docs/adr/0014-standardized-provenance-recording.md
```diff
@@ -0,0 +1,62 @@
...
+
+To do this, we need to specify what provenance information is recorded and how it can be implemented in HERMES to make it easy to use.
+
+## Considered Options
+
+* Provide HERMES API-methods that also document themselves
+
+## Decision Outcome
+
+Chosen option: "Provide HERMES API-methods that also document themselves", because comes out best.
```
sdruskat:
```suggestion
```

As this is in the proposal stage, I think we need more feedback on the actual solution. What are alternatives for this solution? This ADR is already a very good track record of our thinking so far, but I think we need more buy-in before making the actual decision.

That said, this isn't a blocker for the PR, it just says: we need a decision about this further down the line.

docs/adr/0014-standardized-provenance-recording.md
```diff
@@ -0,0 +1,62 @@
...
+    def write(func, path: str, data, *args, **kwargs):
+        # TODO: handle and record byte formats properly
+        with open(path) as fi:
+            func(fi, data, *args, **kwargs)
+        prov.record("write", path, func.__name__, data) # also module of func
+```
+
+* Good, because allows for recording of provenance information of the plugins
+* Good, because it isn't making plugin development harder
+* Bad, because API methods may not cover all I/O functionality python provides
```
sdruskat: True, but we can make a best effort to cover as many as make sense.
When we provide the respective API (`load_file`, `make_request`, etc.)
and make it usable enough so that developers don't see themselves forced to come up with their own solutions,
I think we can get good coverage.

Probably mostly a question of documentation, plugin templates, etc.

docs/adr/hermes-prov-diagram/hermes-prov.svg
sdruskat: As discussed, I think this is a good enough basis for discussion right now. When someone finds some time, they can feel free to create a swimlane version to improve legibility.

zyzzyxdonta:
> That's a good point, but I don't think we're yet clear on the actual schema vocabularies to use.

That's exactly why I said this is way more detailed than I would have done it 😄

zyzzyxdonta:
> As an ADR _draft_ I think this is good to be merged (_iff_ the decision is removed for now). We can postpone the actual decision to before we get to the implementation, which is when we'll have a complete implementation of the plugin steps and merge strategy steps and will be in a better position to find alternatives, complete the picture, prototype options and make a decision.
> 
> Would you agree, @zyzzyxdonta?

Sure 👍🏻

Issue #439, opened 2025-10-06
Ld_list issues


notactuallyfinn:
While testing ld_list in #418 we found some potential issues:

- [x] Sets and graphs are not dictionaries with `"@set"` or `"@graph"` as keys for their list of items in their expanded JSON-LD form. But the constructor of and `is_ld_list` from `ld_list` assume it otherwise (`is_container` of `ld_list` has the same flaws) and the conversion from a ld_list with `container = "@set"` to expanded json gives the wrong form:
  - `{"https://schema.org/name": {"@set": ["a", "b"]}}`'s expanded version is `[{"https://schema.org/name": [{"@value": "a"}, {"@value": "b}]}]`.
  - `is_ld_list([{"@set": ["a", "b"]}])` and `is_ld_list([{"@set": [{"@value": "a"}, {"@value": "b}]}])` return `True` but `is_ld_list([{"@value": "a"}, {"@value": "b}])` returns `False`
  - `ld_list([{"@set": [{"@value": "a"}, {"@value": "b}]}])` returns the correct object but `ld_list([{"@set": ["a", "b"]}])` creates a non expanded ld_list and `ld_list([{"@value": "a"}, {"@value": "b}])` raises a `ValueError`
  - `ld_list.from_list(["a", "b"], container="@set", key="https://schema.org/name", context={"schema": "https://schema.org/"}).ld_value` is `[{"@set": [{"@value": "a"}, {"@value": "b}]}]` instead of `[{"@value": "a"}, {"@value": "b}]}]`
- [ ] Should a list containing a single value (a dictionary) be considered an expanded set/ graph or just as a expanded JSON-LD dict if it isn't a value or id? This is relevant for e.g. setting or appending such an object (for example `[{"schema:name": "foo"}]`) but also for `ld_dict`.
- [x] JSON-LD supports nested containers. But the type of those makes a difference which is not considered in `__setitem__` and `append` of `ld_list`. For example:
  - A set inside a set gets unpacked but a set inside a list gets turned into a list when expanded.
  - `[{"https://schema.org/name": [{"@set": [{"@value": "a"}, {"@set": ["b", "c"]}, {"@list": ["d", "e"]}]}]}]`'s expanded from is `[{"https://schema.org/name": [{"@value": "a"}, {"@value": "b"}, {"@value": "c"}, {"@list": [{"@value": "d"}, {"@value": "e"}]}]}]`
[...]

PR #445, opened 2025-10-17
Change ld_list according to #439

notactuallyfinn:
#439 is not yet completely discussed/ the propositions not approved.
This PR contains the changes that seem plausible as of now.

### DIFF
--- a/src/hermes/model/types/ld_container.py
+++ b/src/hermes/model/types/ld_container.py
@@ -224,7 +224,7 @@ class ld_container:
         :param value: The value that is to be expanded. Different types are expected based on the type of self:
             <ul><li>If type(self) == ld_dict: value must be a dict</li>
             <li>If type(self) == ld_list: value must be a list</li></ul>
-            value will be exapnded as if it was the data_dict/ the item_list of self.
+            value will be expanded as if it was the data_dict/ the item_list of self.
         :type value: JSON_LD_VALUE
 
         :return: The expanded version of value i.e. the data_dict/ item_list of self if it had been value.
@@ -293,14 +293,8 @@ class ld_container:
         for index in range(len(path) - 1, 0, -1):
             current_data = current_data[path[index]]
         # replace the data_dict/ item_list so that value is now inside of the ld_value of parent and store the old value
-        if current_data == []:
-            # itemlist of an empty ld_list:
-            # The item_list can't be replaced like in all other cases
-            self_data = None
-            current_data.append(value)
-        else:
-            self_data = current_data[path[0]]
-            current_data[path[0]] = value
+        self_data = current_data[path[0]]
+        current_data[path[0]] = value
 
         # expand the ld_value of parent to implicitly expand value
         # important the ld_value of parent is not modified because the processor makes a deep copy
@@ -310,10 +304,7 @@ class ld_container:
         )
 
         # restore the data_dict/ item_list to its former state
-        if self_data is not None:
-            current_data[path[0]] = self_data
-        else:
-            current_data.clear()
+        current_data[path[0]] = self_data
 
         # use the path to get the expansion of value
         for index in range(len(path) - 1, -1, -1):
--- a/src/hermes/model/types/ld_list.py
+++ b/src/hermes/model/types/ld_list.py
@@ -59,8 +59,12 @@ class ld_list(ld_container):
         :return:
         :rtype: None
 
-        :raises ValueError: bla
-        :raises ValueError: bla
+        :raises ValueError: If the given key is not a string or None was given.
+        :raises ValueError: If the given data is not a list.
+        :raises ValueError: If the data represents an unexpanded @set. I.e. is of the form [{"@set": [...]}]
+        :raises ValueError: If the given key is "@type" but the container_type not "@set"
+            or a value in the item_list not a string.
+        :raises ValueError: If the given key is not "@type" and any value in the item_list not a dict.
         """
         # check for validity of data
         if not isinstance(key, str):
@@ -272,8 +276,16 @@ class ld_list(ld_container):
             other = [other]
         if isinstance(other, list):
             if ld_list.is_ld_list(other):
-                other = ld_list.get_item_list_from_container(other[0])
-            other = self.from_list(other, parent=self.parent, key=self.key, context=self.context)
+                if "@list" in other[0]:
+                    cont = "@list"
+                elif "@graph" in other[0]:
+                    cont = "@graph"
+                else:
+                    cont = "@set"
+                other = other[0][cont]
+            else:
+                cont = "@set"
+            other = self.from_list(other, parent=self.parent, key=self.key, context=self.context, container_type=cont)
 
         # check if the length matches
         if len(self.item_list) != len(other.item_list):
@@ -338,7 +350,7 @@ class ld_list(ld_container):
                         # swap order if first try returned NotImplemented
                         res = other_item.__eq__(item)
                     # if one of both comparisons returned true the elements are equal
-                    if res:
+                    if res is not NotImplemented and res:
                         equality_pairs[index] += [other_index]
                 if len(equality_pairs[index]) == 0:
                     # there exists no element in other that is equal to item
@@ -624,7 +636,7 @@ class ld_list(ld_container):
         :type value: list[JSON_LD_VALUE | BASIC_TYPE | TIME_TYPE]
         :param parent: The parent container of the new ld_list.<br>If value is assimilated by parent druing JSON-LD
             expansion parent is extended by value and parent is returned.
-        :type parent: ls_container | None
+        :type parent: ld_container | None
         :param key: The key into the inner most parent container representing a dict of the new ld_list.
         :type: key: str | None
         :param context: The context for the new list (is will also inherit the context of parent).<br>
--- a/test/hermes_test/model/types/test_ld_container.py
+++ b/test/hermes_test/model/types/test_ld_container.py
@@ -6,7 +6,7 @@
 # SPDX-FileContributor: Michael Meinel
 # SPDX-FileContributor: Michael Fritzsche
 
-from datetime import datetime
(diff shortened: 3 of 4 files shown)


src/hermes/model/types/__init__.py
```diff
@@ -21,56 +21,48 @@
...
+    (ld_dict.is_json_dict, {"ld_container": lambda c, **kw: ld_dict([c], **kw)}),
+    (lambda v: isinstance(v, str), {"python": lambda v, parent, **_: parent.ld_proc.compact_iri(parent.active_ctx, v)}),
 
     # Convert internal data types to expanded_json
-    (lambda c: ld_container.is_json_id(c) or ld_container.is_json_value(c), dict(expanded_json=lambda c, **_: [c])),
-    (ld_dict.is_json_dict, dict(expanded_json=lambda c, **kw: ld_dict.from_dict(c, **kw).ld_value)),
-    (ld_dict.is_ld_dict, dict(expanded_json=lambda c, **kw: ld_dict.from_dict(c[0], **kw).ld_value)),
+    (lambda c: ld_container.is_json_id(c), {"expanded_json": lambda c, **_: c}),
+    (lambda c: ld_container.is_ld_id(c), {"expanded_json": lambda c, **_: c[0]}),
+    (lambda c: ld_container.is_json_value(c), {"expanded_json": lambda c, **_: [c]}),
```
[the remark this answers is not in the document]
led02: In fact, I've tried both and liked None of them (with the `dict(...)` variant being a bit more readable ... but that's subjective).

If there is a refactoring going on here, do it properly (i.e., providing some class that encapuslates these rules and maybe also allow for better configurability, extensibility, documentabilty, ...). I would really appreciate this, making the connection more explicit without adding to much complexity.

Also note that the introduction of a rule class or similar is most probably required anyways when it comes to configuring the *process* step.

src/hermes/model/types/__init__.py
```diff
@@ -21,56 +21,48 @@
...
+    (lambda c: ld_container.is_ld_value(c), {"expanded_json": lambda c, **_: c}),
```
led02: Why is the callback wrapped in an additional `lambda c` here?

However, this seems like a reasonable and important change... maybe an additinal hint in the comment about the side effects of the ordering of especially these rules might be helpful. (I'll also have look into the commit history to better understand the reasons behind this).

src/hermes/model/types/ld_list.py
```diff
@@ -44,15 +29,7 @@ def __getitem__(self, index):
...
-        # set your_ld_list[index] to the list [{"@type": "foo", "name": "bar"}] given in non expanded form  or
-        # set your_ld_list[index] to the set [{"@type": "foo", "name": "bar"}] given in expanded form
-        #   (ld_list.fromlist([{"@type": "foo", "name": "bar"}]) defaults to container type list
-        #    which would have the object as an expanded form whereas the expanded form of a list would be
-        #    ["@list": [{"@type": "foo", "name": "bar"}]]
-        #    This is relevent because nested sets get unnested when being expanded and lists not.
-        #    Moreover a set inside a list gets automaticaly converted to a list when expanded)
-
-        # FIXME: #439 what happens when a ld_list is put inside another also depends on their container types
+        # set your_ld_list[index] to the list [{"@type": "foo", "name": "bar"}] given in non expanded form
```
led02: Well, it should be considered whether this operation should be allowed in this context. However, in general, it should behave as expected (i.e., set the value to what is passed). What it clearly must not do is setting something in the `data` property that is not valid expanded JSON. People who really want to do this still can directly access item_list or data...

For this concrete question I see two valid solutions: Do handling as to what is easiest implementable (and document) or if in doubt, raise an error.

src/hermes/model/types/__init__.py
```diff
@@ -21,56 +21,48 @@
...
     ),
 
     # Wrap item from ld_dict in ld_list
-    (ld_list.is_ld_list, dict(ld_container=ld_list)),
-    (
-     lambda c: isinstance(c, list) and all(isinstance(item, dict) for item in c),
-     dict(ld_container=lambda c, **kw: ld_list([{"@list": c}], **kw))
-    ),
+    (ld_list.is_ld_list, {"ld_container": ld_list}),
+    (lambda c: isinstance(c, list), {"ld_container": lambda c, **kw: ld_list([{"@list": c}], **kw)}),
```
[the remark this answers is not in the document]
notactuallyfinn: When getting a value from an ld_dict it should return an ld_list always. Therefor it is important to know wheter a given value comes from an ld_list or an ld_dict. If the value, where expanded JSON-LD is expected, is not inside a list it previously was inside a list.
I removed the condition because when getting "@type" from `{"@type": ["A"]}` the value in the list is a string and not a dict. (I found no cases where this goes wrong because everything is valid expanded JSON-LD.)

We should know that the values are valid expanded JSON-LD here because this conversion happens only when converting to the pythonized version of a value that was previously expanded.

src/hermes/model/types/ld_container.py
```diff
@@ -83,13 +83,6 @@ def ld_value(self):
     def _to_python(self, full_iri, ld_value):
         if full_iri == "@id":
             value = self.ld_proc.compact_iri(self.active_ctx, ld_value, vocab=False)
-        elif full_iri == "@type":
-            value = [
-                self.ld_proc.compact_iri(self.active_ctx, ld_type)
-                for ld_type in ld_value
-            ]
-            if len(value) == 1:
-                value = value[0]
```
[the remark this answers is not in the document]
led02: see hint above... this will break conversion of single type classes which is valid but which was already critizised towards me (or maybe there were even real errors... but I doubt it)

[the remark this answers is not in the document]
notactuallyfinn: I haven't found anything it breaks yet, e.g. no failing tests, but it took some effort to get this running.

src/hermes/model/types/ld_container.py
```diff
@@ -41,64 +101,228 @@ def __init__(self, data, *, parent=None, key=None, index=None, context=None):
...
+        to generate a minimal dict.
+
+        The result of this function is what value has turned into.
+
+        :param self: The ld_dict or ld_list in which value gets expanded
+        :type self: Self
+        :param value: The value that is to be expanded. Different types are expected based on the type of self:
+            <ul><li>If type(self) == ld_dict: value must be a dict</li>
+            <li>If type(self) == ld_list: value must be a list</li></ul>
+            value will be exapnded as if it was the data_dict/ the item_list of self.
```
SKernchen:
```suggestion
            value will be expanded as if it was the data_dict/ the item_list of self.
```

src/hermes/model/types/ld_container.py
```diff
@@ -41,64 +101,228 @@ def __init__(self, data, *, parent=None, key=None, index=None, context=None):
...
+                path.extend(["@graph", 0])
+            path.append(self.ld_proc.expand_iri(parent.active_ctx, parent.key) if self.index is None else self.index)
+            if parent.parent is None:
+                break
+            parent = parent.parent
+
+        # if neither self nor any of its parents is a ld_dict:
+        # create a dict with the key of the outer most parent of self and this parents ld_value as a value
+        # this dict is stored in an ld_container and simulates the most minimal JSON-LD object possible
+        if parent.__class__.__name__ != "ld_dict":
```
SKernchen: Then you could append this above

src/hermes/model/types/ld_list.py
```diff
@@ -3,124 +3,515 @@
...
-        if isinstance(other, ld_list):
-            # FIXME: #439 When are ld_lists equal?
-            return self.item_list == other.item_list and self.container == other.container
+        :return: Whether or not value is being considered to be contained in self.
+        :rtype: bool
+        """
+        # expand value
+        expanded_value = self._to_expanded_json([value])
+        # empty list -> no value to check
+        if len(expanded_value) == 0:
```
SKernchen: Is it that way with normal contains?

src/hermes/model/types/ld_list.py
```diff
@@ -3,124 +3,515 @@
...
+        Creates a ld_list from the given list with the given parent, key, context and container_type.<br>
+        Note that only container_type '@set' is valid for key '@type'.<br>
+        Further more note that if parent would assimilate the values in value no new ld_list is created
+        and the given values are appended to parent instead and parent is returned.
+
+        :param value: The list of values the ld_list should be created from.
+        :type value: list[JSON_LD_VALUE | BASIC_TYPE | TIME_TYPE]
+        :param parent: The parent container of the new ld_list.<br>If value is assimilated by parent druing JSON-LD
+            expansion parent is extended by value and parent is returned.
+        :type parent: ls_container | None
```
SKernchen:
```suggestion
        :type parent: ld_container | None
```

src/hermes/model/types/ld_container.py
```diff
@@ -41,64 +101,228 @@ def __init__(self, data, *, parent=None, key=None, index=None, context=None):
...
+            The return type is based on the type of self:
+            <ul><li>If type(self) == ld_dict: the returned values type is dict</li>
+            <li>If type(self) == ld_list: the returned values type is list</li></ul>
+        :rtype: EXPANDED_JSON_LD_VALUE | dict[str, EXPANDED_JSON_LD_VALUE]
+        """
+        # search for an ld_dict that is either self or the inner most parents parent of self that is an ld_dict
+        # while searching build a path such that it leads from the found ld_dicts ld_value to selfs data_dict/ item_list
+        parent = self
+        path = []
+        while parent.__class__.__name__ != "ld_dict":
```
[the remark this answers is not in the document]
notactuallyfinn: I am not sure.
It could work and I see no real reason why it shouldn't but I preferred to use an actually existent ld_dict and not a dummy construct.
What doing it like I implemented would allow us to do though is that we could enforce that every ld_container has an index __xor__ a key (which then actually is the value used the get this object in the parent container).

PR #456, opened 2025-11-21
Fix #454: Add end-to-end tests for the plugin-related `SoftwareMetadata` API

### LINKED ISSUE #454: Create end-to-end tests for the `SoftwareMetadata` API
These should cover ideally all use cases of interaction of plugins with the data model via the API.

sdruskat:


test/hermes_test/model/test_api_e2e.py
```diff
@@ -0,0 +1,355 @@
...
+    config_file = tmp_path / "hermes.toml"
+    config_file.write_text("[harvest]\nsources = [ \"cff\" ]")
+
+    orig_argv = sys.argv[:]
+    sys.argv = ["hermes", "harvest", "--path", str(tmp_path), "--config", str(config_file)]
+    result = {}
+    try:
+        monkeypatch.setattr(context_manager.HermesContext.__init__, "__defaults__", (tmp_path.cwd(),))
+        cli.main()
+    except SystemExit:
```
SKernchen: I don't like the use of SystemExit, can't we just use the code directly and skip the try

test/hermes_test/model/test_api_e2e.py
```diff
@@ -0,0 +1,355 @@
...
+    config_file = tmp_path / "hermes.toml"
+    config_file.write_text("[harvest]\nsources = [ \"codemeta\" ]")
+
+    orig_argv = sys.argv[:]
+    sys.argv = ["hermes", "harvest", "--path", str(tmp_path), "--config", str(config_file)]
+    result = {}
+    try:
+        monkeypatch.setattr(context_manager.HermesContext.__init__, "__defaults__", (tmp_path.cwd(),))
+        cli.main()
+    except SystemExit:
```
SKernchen: Same as above

test/hermes_test/model/test_api_e2e.py
notactuallyfinn: Not really, cli.main() always calls sys.exit().
On feature/454-e2e-deposit I've changed the try-except-finally to read:
```python
try:
    monkeypatch.setattr(context_manager.HermesContext.__init__, "__defaults__", (tmp_path.cwd(),))
    cli.main()
except SystemExit as e:
    if e.code != 0:
        raise e
finally:
    # validation of results
```
I hope that this is better.

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
 # Write a plugin for HERMES
  
 
-This tutorial will present the basic steps for writing an additional harvester.
-At the moment only the architecture for harvester plugins is stable.
-The full code and structure is available at  [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
+This tutorial will present the basic steps for writing additional plugins.
+
+The full code and structure of a harvest plugin is available at [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
```
SKernchen, line 19: Made an Issue to update the plugin for later: https://github.com/softwarepub/hermes-plugin-git/issues/3

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
-In the `hermes.toml` this would be achieved by [harvest.{plugin_name}].
-The `GitHarvestSettings` are associated with the `GitHarvestPlugin`.
-In the plugin you need to overwrite the `__call__` method.
-For now a simple "Hello World" will do. The method returns two dictionaries.
-These will contain the harvested data in CodeMeta (JSON-LD) and additional information, e.g., to provide provenance information.
-That is the basic structure for the plugins source code.
-
-To integrate this code, you have to register it as a plugin in the `pyproject.toml`.
+
+The `__call__` method of harest plugins needs to return a SoftwareMetadata object containing the harvested metadata.
```
SKernchen:
```suggestion
The `__call__` method of harest plugins needs to return a **SoftwareMetadata** object containing the harvested metadata.
```
Or something else to highlight SoftwareMetadata

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
+strategies = {
+    full_type_iri: {
+        full_property_iri: Reject(),
+        ...
+    },
+    ...
+}
+```
+
+HERMES would use the `Reject` strategy for merging values of the key `full_property_iri` in objects of type `full_type_iri`. (A key in strategies being `None` instead of a string indicates to HERMES that its value is to be used as a default [i.e. if no more specific entry exists].)
```
SKernchen: Maybe add link to Reject strategy

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
+
+    def __call__(self, command: HermesCurateCommand, metadata: SoftwareMetadata) -> SoftwareMetadata:
+        data = SoftwareMetadata()
+
+        # TODO: curate the metadata and write it into data
+
+        return data
+```
+
+The `__call__` method of harest plugins needs to return a SoftwareMetadata object containing the curated metadata.
```
SKernchen:
```suggestion
The `__call__` method of curate plugins needs to return a SoftwareMetadata object containing the curated metadata.
```

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -113,55 +313,74 @@ This variant is used to contribute to the HERMES community or adapt the HERMES w
...
+```
+<br><br>
 ```{admonition} Congratulations!
 You can now write plugins for HERMES.
+Consider publishing it for others to use following this guide. TODO: add link
 ```
-To fill the plugin with code, you can check our [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git) repository.
-There is the code to check the local git history and extract contributors of the given branch.
 
 If you have any questions, wishes or requests, feel free to contact us.
```
SKernchen, line 388: In the future we should add a link/information about the plugin marketplace here.

src/hermes/commands/deposit/file.py
```diff
@@ -11,22 +11,23 @@
 from pydantic import BaseModel
 
 from hermes.commands.deposit.base import BaseDepositPlugin
-from hermes.model.path import ContextPath
 
 
 class FileDepositSettings(BaseModel):
-    filename: str = 'hermes.json'
+    filename: str = 'codemeta.json'
```
SKernchen: Maybe someone does not want to update the codemeta.json and just want a screenshot with all the harvestet metadata. So I like hermes.json more or deposit.json?

src/hermes/commands/deposit/invenio_rdm.py
```diff
@@ -47,6 +46,12 @@ def resolve_license_id(self, license_url: t.Optional[str]) -> t.Optional[dict]:
         except HTTPError:
             pass
 
+        # FIXME: Why not get all license_cross_refs and then use a query parameter like this:
+        # ?q=props.url:("license_url" OR "license_cross_ref[1]" OR ...)&size=1000
+        # That would be able to replace _search_license_info.
+        # FIXME: Some licenses in valid_licenses["hits"]["hits"]["props"]["url"] are only http although
+        # https://spdx.org/licenses/license.json lists them in crossRef as https
+
```
SKernchen, line 54: Yes that should be possible.
We should fix the licenses to https.

src/hermes/commands/deposit/invenio_rdm.py
```diff
@@ -65,6 +70,7 @@ def resolve_license_id(self, license_url: t.Optional[str]) -> t.Optional[dict]:
                 if license_info is not None:
                     break
             else:
+                # FIXME: Why is this only raised here and not always when license_info is None?
```
SKernchen, line 73: I think, it was originally because that was were the deposit step broke two years ago, when Zenodo implemented the need of the correct license identifier. But you are right, we should fix that.

src/hermes/commands/postprocess/invenio.py
```diff
@@ -3,47 +3,73 @@
...
 
 
-def cff_doi(ctx):
-    deposition_path = ctx.get_cache('deposit', 'deposit')
-    with deposition_path.open("r") as deposition_file:
-        deposition = json.load(deposition_file)
-    try:
-        cff = yaml.load(open('CITATION.cff', 'r'), yaml.Loader)
-        new_identifier = {
+class cff_doi(HermesPostprocessPlugin):
```
SKernchen, line 48: Maybe also for codemeta, because we can not be sure the user has a cff

src/hermes/commands/process/standard_merge.py
```diff
@@ -0,0 +1,850 @@
...
+
+ACTIONS = {
+    "default": MergeSet(DEFAULT_MATCH),
+    "concat": Concat(),
+    "Person": MergeSet(MATCH_FUNCTION_FOR_TYPE["schema:Person"]),
+    **{
+        "Or".join(types): MergeSet(match_multiple_types(
+            *(("schema:" + type, MATCH_FUNCTION_FOR_TYPE.get("schema:" + type, DEFAULT_MATCH)) for type in types)
+        ))
+        for types in [
```
SKernchen, line 173: Maybe we can curl this information? Or automate the mapping?
But really good work 👍

docs/source/tutorials/writing-a-plugin-for-hermes.md
notactuallyfinn: I made it a link, which is highlighted by default.

docs/source/tutorials/writing-a-plugin-for-hermes.md
notactuallyfinn: Done

docs/source/tutorials/writing-a-plugin-for-hermes.md
notactuallyfinn: Done

docs/source/tutorials/writing-a-plugin-for-hermes.md
notactuallyfinn: Added a link

src/hermes/commands/deposit/file.py
notactuallyfinn: That is something I didn't consider before.
I agree and changed it.

test/hermes_test/commands/deposit/test_invenio_e2e.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+
+from hermes.commands import cli
+from hermes.model import context_manager
+from hermes.model.api import SoftwareMetadata
+
+
+@pytest.fixture
+def sandbox_auth():
+    path = Path("./../auth.txt")
```
SKernchen: We need to tell users how to provide auth or skip the test for ci

SKernchen: Especially for our testsuite ci

test/hermes_test/commands/deposit/test_invenio_e2e.py
```diff
@@ -0,0 +1,97 @@
...
+    if not path.exists():
+        pytest.skip("Local auth token file does not exist.")
+    with path.open() as f:
+        yield f.read()
+
+
+@pytest.mark.parametrize(
+    "metadata, invenio_metadata",
```
SKernchen, line 26: invenio_metada can change, maybe we should test the format as well?

test/hermes_test/commands/process/test_process.py
SKernchen, line 1: Maybe also test for 3-way-merges. And test strategies in different files would be great.

test/hermes_test/commands/deposit/test_invenio_e2e.py
notactuallyfinn: Yes, implementing a way to provide auth seems reasonable.
(But the test is already skipped if the file doesn't exist.)

src/hermes/commands/process/standard_merge.py
notactuallyfinn: Maybe.
I'll think about whether or not this can be (easily) done.
Edit: It seems doable, I'll implement it in the next days.

SKernchen:
Findings from user test (tested in poetry environment on command line):
process step:
 - process step fails when one harvester failed in the harvest step
 (maybe process should not look into toml?)

deposit step:
error:
 - 'ld_list' object has no attribute 'startswith'

 - for file add log.info where to find file would be good

notactuallyfinn:
Could you please provide more information on this error?
> 'ld_list' object has no attribute 'startswith'

src/hermes/commands/deposit/invenio_rdm.py
SKernchen:
```suggestion
class InvenioRDMDepositPlugin(InvenioDepositPlugin):
```

pyproject.toml
SKernchen:
```suggestion
invenio_rdm = "hermes.commands.deposit.invenio_rdm:InvenioRDMDepositPlugin"
```

SKernchen:
> Could you please provide more information on this error?
> 
> > 'ld_list' object has no attribute 'startswith'

That is the only information provided for this error:
`
1774021073.908437:          hermes.cli:     ERROR | An error occurred during execution of deposit (Find details in './hermes.log')
1774021073.910363:          hermes.cli:     DEBUG | Original exception was: 'ld_list' object has no attribute 'startswith'
`
I do not know where it is coming from. Everything with startswith seemed fine to me 🤷‍♀️

src/hermes/commands/deposit/invenio_rdm.py
notactuallyfinn: Done

pyproject.toml
notactuallyfinn: Done

notactuallyfinn:
Okay, I think I found the error and hopefully fixed it. Please let me know if the commit didn't fix it.
(Can't verify, I never ran into that error.)

SKernchen:
Yes, this error is gone. Now I'm getting
`1774263836.359856:          hermes.cli:     DEBUG | Original exception was: The given license in CodeMeta must be of type str. Licenses of type 'CreativeWork' are not supported.`
Though the license is correct.

src/hermes/commands/deposit/base.py
```diff
@@ -125,29 +135,24 @@ def init_command_parser(self, command_parser: argparse.ArgumentParser) -> None:
...
+        # run plugin
         try:
             plugin_func(self)
-
         except HermesValidationError as e:
-            self.log.error("Error while executing %s: %s", plugin_name, e)
-            self.errors.append(e)
+            self.log.error(f"Error while executing {plugin_name}: {e}")
+            raise HermesPluginRunError(
+                f"Something went wrong while running the curate plugin {self.settings.plugin}"
```
SKernchen:
```suggestion
                f"Something went wrong while running the deposit plugin {self.settings.plugin}"
```

SKernchen:
> Yes, this error is gone. Now I'm getting `1774263836.359856: hermes.cli: DEBUG | Original exception was: The given license in CodeMeta must be of type str. Licenses of type 'CreativeWork' are not supported.` Though the license is correct.

Ok so this error comes from invenio.py line 216:
```
        if not isinstance(license_url, str):
            raise RuntimeError(
                "The given license in CodeMeta must be of type str. "
                "Licenses of type 'CreativeWork' are not supported."
            )
```
where license_url is in my case {"@id": "{license}"}, so this error triggers

SKernchen:
After giving the correct input (using license_url["@id]) it worked, but if your case has different input, we should find the source for this problem.

notactuallyfinn:
According to [schema](https://schema.org/license) the values for the license property are URLs or CreativeWork objects.
You passed the URL as `{"@id": your_url}` which is not a value for URLs (that should be `{"@value": your_url}`, (shouldn't it?), which then gets replaced by `your_url` by the ld_container). Therefore `{"@id": your_url}` is handled as a value of type CreativeWork, which is rejected because the conversion is non-trivial.
That is the reason for the new error.
I'll add an exemption for objects of this format `{"@id": your_url}` for now.
(I'm aware that in tests the format `{"@id": your_url}` is required. That should/ will be changed.)

SKernchen:
Yeah but why is it passed like that, is the real question right.
```
"http://schema.org/license": [{"@id": "https://spdx.org/licenses/Apache-2.0"}]
```
is it in the expanded.json from the curate step

notactuallyfinn:
I don't know why this was (and still is) a requirement for the harvest and process steps.
It worked previously because `{"@id": value}` was returned by ld_dicts as `value`, which is not good (the information that this is an id is lost). This was changed so that now an ld_dict with this data_dict (`{"@id": value}`) is returned.
But that leads to the problem that `{"@id": your_url}` is no longer returned as a string leading to the error you saw.
I think the options now are to either always try if the id of a CreativeWork object is a valid url for a license or just  disallow the use of CreativeWork objects and expect the license to be passed as a value of a URL (i.e. `{"@value": your_url}`).

led02:
> It worked previously because `{"@id": value}` was returned by ld_dicts as `value`, which is not good (the information that this is an id is lost).

This information about the license being an ID is actually contained in the codemeta context:  https://github.com/codemeta/codemeta/blob/13bb230d250775026691a102a24b97dea7669bf6/codemeta.jsonld#L44

test/hermes_test/commands/deposit/test_invenio_e2e.py
notactuallyfinn: I'm not quite sure what you mean.
Isn't the format tested indirectly?

SKernchen:
Nice now it works:
More points I found:
- With the postprocess plugin (invenio_rdm) in hermes.toml removes license header and sorts postprocess above deposit (and add "," to every list)
- harvest: breaks if one plugin is not found and don't run the others (if broken plugin is first)
- process broke with `'str' object has no attribute 'append'` with a codemeta with conflicts in it
- deposit without --initial broke with `Error while executing invenio_rdm: Found two different record ids or conflicting metadata.`

test/hermes_test/commands/deposit/test_invenio_e2e.py
notactuallyfinn: It should now be possible to pass the auth token when running the tests.
To do this pass the custom command line option `--sandbox_auth your_auth_token` to the pytest command.

We will have to get a ZenodoSanbox account and auth token to put it into the GitHub secrets and then adapt the test workflow.

test/hermes_test/commands/deposit/test_invenio_e2e.py
SKernchen: Yes thats true, than we can leave it like that

test/hermes_test/commands/process/test_process.py
notactuallyfinn: Do the test now are enough/ cover everything you would like to be covered?

PR #463, opened 2026-01-09
Feature/457 good init no dataverse

nheeb:
This includes all the recent init updates without the dataverse elements.
Those init updates roughly include
- using tags to trigger the pipeline
- using jinja for ci templates
- installing hermes plugins with init
- using comandline params to choose different hermes versions
- supporting different deposit platforms
- better usability
- bug fixes
- big warning when using outdated hermes version


poikilotherm:
We should merge these in "squash" mode to avoid having the Dataverse bits in the commit history. The back and forth merging already seems to have caused some trouble.

I played hermes init on the showcase repo and found mixed results when aborting. It claims "no changes were made", while actually there were some:

grafik

nheeb:
Reminder: Also merge "transition branch" in the ci-templates repo when merging this

poikilotherm (approved):
Reviewed and tested locally. Good to go from my side! Ship it 🚢

PR #473, opened 2026-02-20
Feature/harvesting metadata from a provided repository url via cloning

Aidajafarbigloo:
This feature branch introduces functionality for `harvesting metadata from a provided repository URL via cloning`.

Changes:
1. Accept a repository **URL** as a parameter in the `hermes harvest` command
2. Accept a **token** (for GitHub/GitLab) as a parameter in the `hermes harvest` command (for the plugin githublab)
3. Clone the repository locally
4. Harvest metadata from the cloned repository (CFF and CodeMeta)

### DIFF
--- a/src/hermes/commands/clean/base.py
+++ b/src/hermes/commands/clean/base.py
@@ -6,6 +6,7 @@
 
 import argparse
 import shutil
+import logging
 
 from pydantic import BaseModel
 
@@ -25,6 +26,7 @@ class HermesCleanCommand(HermesCommand):
 
     def __call__(self, args: argparse.Namespace) -> None:
         self.log.info("Removing HERMES caches...")
+        logging.shutdown()
 
         # Naive implementation for now... check errors, validate directory, don't construct the path ourselves, etc.
         shutil.rmtree(args.path / '.hermes')
--- a/src/hermes/commands/harvest/util/clone.py
+++ b/src/hermes/commands/harvest/util/clone.py
@@ -19,29 +19,62 @@ from typing import Sequence
 # ---------------- utilities ----------------
 
 def _normalize_clone_url(url: str) -> str:
+    """
+    Normalize a repository "clone target" into a format that `git clone` accepts.
+
+    Supported inputs:
+      - SSH scp-like form:      git@host:group/repo(.git)
+      - HTTPS URLs:             https://host/group/repo(.git)
+
+    Normalization rules:
+      - For SSH and HTTPS, append ".git" when missing (common, but not required by all hosts).
+    """
     s = str(url).strip()
+    
+    # SSH scp-style: git@github.com:org/repo
     if re.match(r'^[\w.-]+@[\w.-]+:.*', s):
         return s if s.endswith('.git') else s + '.git'
+    
+    # file:// URLs should be passed as-is.
     if s.startswith('file://'):
         return s
+    
+    # If it's an existing local path
     if os.path.exists(s):
         return s
+    
+    # Parse normal URLs (http/https).
     p = urlparse(s)
     if p.scheme in ('http', 'https'):
         path = p.path if p.path.endswith('.git') else (p.path.rstrip('/') + '.git')
         return f"{p.scheme}://{p.netloc}{path}"
+    
+    # If the user already provided a .git suffix but it isn't http/https, accept it as-is.
     if s.endswith('.git'):
         return s
     raise ValueError(f"Unsupported repository URL format: {url!r}")
 
 def _clear_readonly(func, path, excinfo):
     """
-    onerror handler for shutil.rmtree: try to remove read-only flag and retry.
+    Error handler for `shutil.rmtree(..., onerror=...)`.
+
+    Purpose:
+      Some platforms/tools (notably Windows, antivirus scanners, or git itself) can leave files
+      marked read-only, causing deletion failures. This handler attempts to:
+        1) Remove the read-only attribute, then
+        2) Retry removal of the file/directory.
+
+    Parameters:
+      func:    The function that raised the exception (provided by shutil).
+      path:    The filesystem path that couldn't be removed.
+      excinfo: Exception info tuple (type, value, traceback).
     """
+    # make the path writable.
     try:
         os.chmod(path, stat.S_IWRITE)
     except Exception:
         pass
+    # retry deletion.
     try:
         if os.path.isdir(path):
             shutil.rmtree(path)
@@ -52,10 +85,22 @@ def _clear_readonly(func, path, excinfo):
 
 def rmtree_with_retries(path: Path, retries: int = 6, initial_wait: float = 0.1):
     """
-    Best-effort removal of path with retries and read-only handling.
-    - retries: number of attempts
-    - initial_wait: initial sleep (multiplies by 2 each retry)
-    Logs exceptions but never raises.
+    Recursive directory deletion with retries and read-only handling, for environments where temporary directories may be locked
+    or marked read-only (e.g., Windows, CI systems, antivirus interference).
+
+    Behavior:
+      - If `path` doesn't exist: return immediately.
+      - Attempts deletion up to `retries` times.
+      - Between attempts, sleeps with exponential backoff:
+          wait = initial_wait, then wait *= 2 each retry.
+      - Makes files/directories writable before trying `shutil.rmtree`.
+      - Uses `_clear_readonly` for additional resilience.
+      - Never raises: logs warnings/errors and returns.
+
+    Parameters:
+      path:         Directory to remove.
+      retries:      Number of attempts before giving up.
+      initial_wait: Starting sleep duration (seconds) for exponential backoff.
     """
     if not path.exists():
         return
@@ -80,6 +125,7 @@ def rmtree_with_retries(path: Path, retries: int = 6, initial_wait: float = 0.1)
 
             shutil.rmtree(path, onerror=_clear_readonly)
             
+            # If deletion succeeded, stop.
             if not path.exists():
                 return
         except Exception as e:
@@ -98,14 +144,31 @@ def rmtree_with_retries(path: Path, retries: int = 6, initial_wait: float = 0.1)
     except Exception:
         pass
 
+    # If still present, report and exit without raising.
     if path.exists():
         print(f"error: failed to remove temp dir {path!s} after {retries} attempts. "
               f"Please remove it manually. (Often caused by antivirus or open handles.)")
 
 def _move_or_copy(src: Path, dst: Path):
+    """
+    Move a directory into place, falling back to copy+delete when a move isn't possible.
+
+    Primary strategy:
+      - `os.replace(src, dst)` performs an atomic rename/move when possible.
+
+    Fallback strategy:
+      - If atomic move fails (commonly due to cross-device boundaries or permission issues),
+        copy the directory tree to `dst`, then remove `src` using robust cleanup.
+
+    Parameters:
+      src: Source directory path (typically a temp clone directory).
+      dst: Destination directory path.
+    """
     try:
+        # Fast path: atomic rename (preferred when possible)
         os.replace(str(src), str(dst))
     except Exception:
+        # Cross-device or permission failure — fall back to copy + cleanup
         shutil.copytree(str(src), str(dst))
         rmtree_with_retries(src)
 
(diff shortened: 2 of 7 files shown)


Aidajafarbigloo:
@sferenz 
Could you please take a look at this pull request and share your feedback?

src/hermes/commands/clean/base.py
```diff
@@ -6,6 +6,7 @@
 
 import argparse
 import shutil
+import logging
```
sferenz, line 9: @Aidajafarbigloo Is it necessary for your changes to include logging? If not, please remove it everywhere.

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,249 @@
...
+import tempfile
+import time
+import stat
+from pathlib import Path
+from urllib.parse import urlparse
+from typing import Sequence
+
+# ---------------- utilities ----------------
+
+def _normalize_clone_url(url: str) -> str:
```
sferenz, line 21: Please provide a general comment for each function.

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,249 @@
+# SPDX-FileCopyrightText: 2026 OFFIS e.V.
```
sferenz: Please put UOL here.

src/hermes/commands/harvest/util/token.py
```diff
@@ -0,0 +1,71 @@
...
+
+# SPDX-FileContributor: Stephan Ferenz
+# SPDX-FileContributor: Aida Jafarbigloo
+
+from pathlib import Path
+import base64
+import toml
+
+
+def _load_config(config_path: str) -> dict:
```
sferenz, line 13: Please ensure that every function has a comment

src/hermes/commands/clean/base.py
Aidajafarbigloo: @sferenz Thank you for the comments.
When using `hermes clean` command on Windows, I face this error:
`Run subcommand clean`
`Removing HERMES caches...`
`An error occurred during execution of clean (Find details in './hermes.log')`

Error in the "hermes.log":
`Original exception was: [WinError 32] The process cannot access the file because it is being used by another process: '.hermes\\audit.log'`.

This happens because Windows does not allow deletion of a file that is still open by the current process. The `audit.log` file inside `.hermes` is held open by a logging file handler. When `shutil.rmtree()` attempts to remove the directory, it fails due to the open file handle. I'm using `logging.shutdown()` to ensure that all logging handlers are closed before the directory is deleted, it does not introduce new logging behavior.

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: Sure.

src/hermes/commands/clean/base.py
zyzzyxdonta: This was fixed before but somehow the fix got lost... Weird 🤔 

https://github.com/softwarepub/hermes/pull/226

led02: Well, not only once. But it seems more important to not have a log file in the working directory than having a properly cleaned .hermes cache.

you should be able to configure the path to the logfile, though.

src/hermes/commands/harvest/util/token.py
Aidajafarbigloo: Docstrings were added.

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: Docstrings were added.

Aidajafarbigloo:
@zyzzyxdonta @led02 Could you please review the PR and let me know if something needs to be changed?

src/hermes/tests/bulk_repository_test.py
zyzzyxdonta, line 1: The tests are currently a standalone script but they should use the [Python unittest library](https://docs.python.org/3/library/unittest.html) or [Pytest](https://docs.pytest.org/en/stable/) just like all the other tests do. You rebuilt a lot of things that already exist in these libraries, e.g. output of messages, test fixtures, ...

src/hermes/tests/bulk_repository_test.py
```diff
@@ -0,0 +1,283 @@
...
+- Compares current results to a persisted state file (`hermes_bulk_test_state.json`) to detect regressions
+  (previously "yes" turning into "no").
+- Writes a CSV report (`hermes_bulk_test_results.csv`).
+
+Notes:
+- Tokens are currently set as empty strings, need to be set before use.
+"""
+
+import subprocess
+import pandas as pd
```
zyzzyxdonta, line 24: Pandas is not part of the dependencies and I would prefer not to add it.

src/hermes/tests/bulk_repository_test.py
```diff
@@ -0,0 +1,283 @@
...
+GITLAB_TOKEN = ""
+
+# ---------------- RESULTS STORAGE ----------------
+# In-memory accumulation of per-repository run results.
+results = []
+
+# JSON file that stores previous yes/no answers per URL to detect regressions across runs.
+state_file = "hermes_bulk_test_state.json"
+
+# Load previous state (if any).
```
zyzzyxdonta, line 57: I don't think comparing against previous results makes sense here. Either the code does the right thing (tests succeed) or it doesn't (tests fail). There is no performance degradation over time.

src/hermes/tests/bulk_repository_test.py
```diff
@@ -0,0 +1,283 @@
...
+
+    try:
+        # Step 1: Clean previous metadata
+        subprocess.run(["hermes", "clean"], check=True)
+        print("✅ 'hermes clean' finished.")
+
+        # Step 2: Run harvest for the repository.
+        # Capture stdout/stderr to include HERMES output as an error message when needed.
+        proc = subprocess.run(
+            ["hermes", "harvest", "--url", url, "--token", token],
```
zyzzyxdonta, line 192: I don't think it's a good idea to run the whole program like this. Especially, we shouldn't actually clone any real repositories. Doing this can easily lead to our tests degrading, e.g. because repositories change over time or are deleted.

Here are some things that could be tested instead:

- Git repo URLs are constructed correctly
- git command lines (that are passed to subprocess.run) are constructed correctly

I'm not averse to the idea of having end-to-end tests for the whole metadata extraction workflow. But I think it should be based on a dedicated test repo that should ideally be part of the `tests/` directory and not cloned. It should be safe to assume that a `git clone` works if the command parameters are correct.

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+    Normalize a repository "clone target" into a format that `git clone` accepts.
+
+    Supported inputs:
+      - SSH scp-like form:      git@host:group/repo(.git)
+      - HTTPS URLs:             https://host/group/repo(.git)
+
+    Normalization rules:
+      - For SSH and HTTPS, append ".git" when missing (common, but not required by all hosts).
+    """
+    s = str(url).strip()
```
zyzzyxdonta, line 32: More expressive variable names would be nice. `stripped_url`? Or just overwrite the `url` parameter with the new value so the unstripped value isn't accidentally used again.

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+    # file:// URLs should be passed as-is.
+    if s.startswith('file://'):
+        return s
+    
+    # If it's an existing local path
+    if os.path.exists(s):
+        return s
+    
+    # Parse normal URLs (http/https).
+    p = urlparse(s)
```
zyzzyxdonta, line 47: More expressive variable name, e.g. `parsed_url`

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+        ValueError:   If the repository URL format is invalid.
+    """
+    dest_path = Path(dest_dir)
+    parent = dest_path.parent
+    parent.mkdir(parents=True, exist_ok=True)
+
+    clone_url = _normalize_clone_url(url)
+    
+    # Some GitLab setups have compatibility issues with partial/shallow clones
+    is_gitlab = "gitlab.com" in url.lower()
```
zyzzyxdonta, line 228: See my comments here: https://github.com/softwarepub/hermes/pull/278#pullrequestreview-4013291232

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+    if is_gitlab:
+        if verbose:
+            print("⚠️ GitLab detected: disabling --depth and --filter=blob:none for safety.")
+        depth = None
+        filter_blobs = False
+
+    env = os.environ.copy()
+    if insecure_ssl:
+        # Disables SSL verification (security risk; use only when necessary)
+        env["GIT_SSL_NO_VERIFY"] = "1"
```
zyzzyxdonta, line 238: Why?

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+            pass
+    except Exception:
+        pass
+
+    # If still present, report and exit without raising.
+    if path.exists():
+        print(f"error: failed to remove temp dir {path!s} after {retries} attempts. "
+              f"Please remove it manually. (Often caused by antivirus or open handles.)")
+
+def _move_or_copy(src: Path, dst: Path):
```
zyzzyxdonta, line 152: Why make things this complicated? We don't need atomic moves here.

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+                else:
+                    rmtree_with_retries(dest_path)
+
+            _move_or_copy(tmp2, dest_path)
+            if verbose:
+                print("✅ Repository cloned successfully (fallback/full clone).")
+            return
+
+        # Optimized clone succeeded
+        if dest_path.exists():
```
zyzzyxdonta, line 304: There is no reason to run into this problem at all. Just create a new temporary directory for every clone attempt.

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+    recursive: bool = True,
+    depth: int | None = 1,
+    filter_blobs: bool = True,
+    sparse: bool = False,
+    branch: str | None = None,
+    insecure_ssl: bool = False,
+    *,
+    root_only: bool = False,
+    include_files: Sequence[str] | None = None,
+    verbose: bool = False,
```
zyzzyxdonta, line 189: Instead of this `verbose` flag and `print()` calls, the logging features should be used with appropriate log levels (info, warning, error, ...)

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+            shutil.rmtree(path)
+        else:
+            os.remove(path)
+    except Exception:
+        pass
+
+def rmtree_with_retries(path: Path, retries: int = 6, initial_wait: float = 0.1):
+    """
+    Recursive directory deletion with retries and read-only handling, for environments where temporary directories may be locked
+    or marked read-only (e.g., Windows, CI systems, antivirus interference).
```
zyzzyxdonta, line 89: I wouldn't call this "retries". `git clone` always writes the `.git/` directory as read-only so the issue that something can not be removed will always occur. I think `shutil.rmtree` with the `_clear_readonly` handler should suffice 🤔

src/hermes/commands/harvest/util/clone.py
```diff
@@ -0,0 +1,346 @@
...
+                        pass
+
+            shutil.rmtree(path, onerror=_clear_readonly)
+            
+            # If deletion succeeded, stop.
+            if not path.exists():
+                return
+        except Exception as e:
+            print(f"warn: rmtree attempt {attempt} failed for {path!s}: {e!r}")
+        time.sleep(wait)
```
zyzzyxdonta, line 133: Something is seriously wrong if a sleep is required.

zyzzyxdonta (changes requested):
I left some comments on the cloning procedure and the tests. I haven't looked at how it is integrated into the application yet.

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: Commit 5dd0777

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: Commit 8dca14e

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: Addressed in c0c143b9e3959de843678947045f83af6a70081d

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: Addressed in 8dca14e1fdfe0dbb4f4c95eb0eb9f827225b375c

src/hermes/commands/harvest/util/clone.py
Aidajafarbigloo: I added it as a fallback in case HTTPS cloning failed, I faced the error while testing a few repositories. I removed it in commit 7e21b72.

PR #480, opened 2026-04-29
(Draft) New Data Model

SKernchen:
This pull request is not ready yet. It is only for showing the process


conftest.py
zyzzyxdonta, line 1: Should a top-level conftest exist? I think this belongs into `test/` or `test/hermes_test/`

pyproject.toml
```diff
@@ -38,71 +38,76 @@ dependencies = [
     "pydantic>=2.5.1, <3.0.0",
     "pydantic-settings>=2.1.0, <3.0.0",
     "requests-oauthlib>=2.0.0, <3.0.0",
-    "pynacl>=1.6.2, <2.0.0",
-    "jinja2 (>=3.1.6,<4.0.0)",
+    "pynacl>=1.5.0, <2.0.0",
+    "rdflib (>=7.1.4,<8.0.0)",
+    "schemaorg (>=0.1.1,<0.2.0)",
+    "tomlkit (>=0.14.0,<0.15.0)",
```
zyzzyxdonta: This is our second toml dependency (we already had `toml`. And there is also `tomli` as a transitive dependency. It might make sense to consolidate this into one implementation at some point, if possible.

pyproject.toml
```diff
@@ -38,71 +38,76 @@ dependencies = [
...
 cff_doi = "hermes.commands.postprocess.invenio:cff_doi"
+codemeta_doi = "hermes.commands.postprocess.invenio:codemeta_doi"
 
-[tool.poetry.dependencies]
-python = ">=3.11,<4.0.0"
+[project.entry-points."hermes.process"]
+codemeta = "hermes.commands.process.standard_merge:CodemetaProcessPlugin"
+
+[project.entry-points."hermes.curate"]
+pass_curate = "hermes.commands.curate.pass_curate:DoNothingCuratePlugin"
```
zyzzyxdonta: "do nothing" and "pass" are two different names for the same thing. We should make up our minds and choose one of the two.

zyzzyxdonta (commented):
Some first very high-level comments. Will continue the review another day.

src/hermes/model/hermes_cache.py
zyzzyxdonta, line 1: (Without having read any of the docs,) I find the jargon strange in this module. The module is called `context_manager`. It contains `HermesCache` which is a context manager, and `HermesContext` which is not a context manager but returns context managers. Maybe this could be cleared up by renaming the module. Maybe nothing with `context` in it as this term is already used for the context of the JSON-LD object.

src/hermes/model/hermes_cache.py
```diff
@@ -0,0 +1,198 @@
...
+import os.path
+from pathlib import Path
+from types import TracebackType
+from typing import Optional
+from typing_extensions import Self
+
+from .error import HermesContextError
+
+
+class HermesCache:
```
zyzzyxdonta, line 18: It seems this class is only ever instantiated in `HermesContext`. So it might be a good idea to add a docstring to the module that explains how to get such a cache object (i.e. via `HermesContext`).

src/hermes/model/error.py
```diff
@@ -0,0 +1,72 @@
...
+
+# SPDX-FileContributor: Michael Meinel
+# SPDX-FileContributor: Stephan Druskat <stephan.druskat@dlr.de>
+
+from typing import Any, Union
+
+
+class HermesValidationError(Exception):
+    """
+    This exception should be thrown when input validation (e.g., during harvest) occurs.
```
zyzzyxdonta:
```suggestion
    This exception should be raised when an error occurs during input validation (e.g., during harvest).
```

src/hermes/model/error.py
```diff
@@ -0,0 +1,72 @@
...
+             validate_some_data(src_file)
+        except ValueError as e:
+            raise HermesValidationError(src_file) from e
+    """
+    pass
+
+
+class HermesContextError(Exception):
+    """
+    This exception should be thrown when interacting with the model context.
```
zyzzyxdonta:
```suggestion
    This exception should be raised when interacting with the model context.
```

src/hermes/model/merge/container.py
```diff
@@ -0,0 +1,336 @@
...
+        *,
+        parent: Optional[ld_container] = None,
+        key: Optional[str] = None,
+        index: Optional[int] = None,
+        context: Optional[list[Union[str, JSON_LD_CONTEXT_DICT]]] = None,
+        strategies: dict[Optional[str], dict[Optional[str], MergeAction]] = {}
+    ) -> None:
+        """
+        Create a new ld_merge_list.
+        For further information on this function and the errors it throws see :meth:`ld_list.__init__`.
```
zyzzyxdonta:
```suggestion
        For further information on this function and the exceptions it raises see :meth:`ld_list.__init__`.
```

test/hermes_test/model/types/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+import pytest
+
+
+class MockDocument:
+    """
+    Helper that provides valid JSON-LD data.
+    """
+
+    @classmethod
+    def vocabulary(cls, base_url: str = "http://spam.eggs/") -> dict:
```
zyzzyxdonta: Only example.org and example.com are safe test domains

src/hermes/model/types/ld_container.py
```diff
@@ -0,0 +1,494 @@
...
+]
+""" Type description of an compacted JSON_LD object """
+TIME_TYPE: TypeAlias = Union[datetime, date, time]
+""" All time related types in Python recogniced by ld_Containers """
+JSON_LD_VALUE: TypeAlias = Union[
+    list[Union["JSON_LD_VALUE", BASIC_TYPE, TIME_TYPE, "ld_dict", "ld_list"]],
+    dict[str, Union["JSON_LD_VALUE", BASIC_TYPE, TIME_TYPE, "ld_dict", "ld_list"]],
+]
+""" Type description of valid JSON_LD objects that are partially represented by ld_containers """
+PYTHONIZED_LD_CONTAINER: TypeAlias = Union[
```
zyzzyxdonta: What does pythonized mean? 😅  The Python representation of the thing?

src/hermes/model/types/pyld_util.py
```diff
@@ -0,0 +1,128 @@
...
+        'mappings': {}
+    })
+
+    def expand_iri(self, active_ctx: t.Any, short_iri: str, vocab: bool = True) -> str:
+        return self._expand_iri(active_ctx, short_iri, vocab=vocab)
+
+    def compact_iri(self, active_ctx: t.Any, long_iri: str, vocab: bool = True, value: dict = None) -> str:
+        for mapping in active_ctx["mappings"].values():
+            if "@container" in mapping and long_iri:
+                value = {x: "none" for x in mapping["@container"]}
```
zyzzyxdonta, line 109: I don't understand what this is for.

The `and long_iri` I guess saves work if no IRI is given. But this could be expressed more clearly by doing `if long_iri is None: return long_iri` at the beginning of the function, just like `_compact_iri()` does it.

`value` overwritten in each new iteration in each iteration because it `@container` can only exist once? In that case, this would be clearer by breaking out of the loop.

But what does `"none"` do?

I think this whole module goes deep into pyld internals and a couple of comments would be helpful to understand what is going on.

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
 
-*hermes* uses an internal data model to store the output of the different stages.
-All the data is collected in a directory called `.hermes` located in the root of the project directory.
+`hermes`' internal data model acts like a contract between `hermes` and plugins.
+It is based on [**JSON-LD (JSON Linked Data)**](https://json-ld.org/), and
+the public API simplifies interaction with the data model through Python code.
 
-You should not need to interact with this data directly.
-Instead, use {class}`hermes.model.context.HermesContext` and respective subclasses to access the data in a consistent way.
+Output of the different `hermes` commands consequently is valid JSON-LD, serialized as JSON, that is cached in
```
zyzzyxdonta: >valid JSON-LD, serialized as JSON

Obviously 😁 


```suggestion
Output of the different `hermes` subcommands consequently are valid JSON-LD files that are cached in 
```

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
-*hermes* uses an internal data model to store the output of the different stages.
-All the data is collected in a directory called `.hermes` located in the root of the project directory.
+`hermes`' internal data model acts like a contract between `hermes` and plugins.
+It is based on [**JSON-LD (JSON Linked Data)**](https://json-ld.org/), and
+the public API simplifies interaction with the data model through Python code.
 
-You should not need to interact with this data directly.
-Instead, use {class}`hermes.model.context.HermesContext` and respective subclasses to access the data in a consistent way.
+Output of the different `hermes` commands consequently is valid JSON-LD, serialized as JSON, that is cached in 
+subdirectories of the `.hermes/` directory that is created in the root of the project directory.
```
zyzzyxdonta: Maybe add a little directory tree like this (but updated with current file names):

```
.hermes
├── harvest
│   ├── cff_contexts.json
│   ├── cff.json
│   ├── file_exists_contexts.json
│   └── file_exists.json
└── process
    ├── hermes.json
    └── tags.json
```

(`tree .hermes` in Linux or `tree /f .hermes` in Powershell on Windows)

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
 
+The cache is purely for internal purposes, its data should not be interacted with.
```
zyzzyxdonta: Maybe the other way around:


```suggestion
The cache should only be interacted with via the `hermes` libraries.
```

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
+
+#### How plugins work with the API
+
+```{important}
+Plugins access the data model _exclusively_ through the API class {class}`hermes.model.SoftwareMetadata`.
+```
+ 
+The following sections show how this class works. 
+
+##### Creating a data model instance
```
zyzzyxdonta: h5 is too many levels down. It's absolutely fine to have multiple paragraphs with multiple sentences each in a subdivision.

poetry.lock
zyzzyxdonta, line 1: With the given `poetry.lock`, the docs don't compile. A lot of our packages are old and incompatible with the ones that aren't pinned. I got it to work like this:

```diff
diff --git i/pyproject.toml w/pyproject.toml
index 17ea508..932427b 100644
--- i/pyproject.toml
+++ w/pyproject.toml
@@ -92,13 +92,13 @@ pytest-httpserver = "^1.1.5"
 optional = true
 
 [tool.poetry.group.docs.dependencies]
-Sphinx = "^6.2.1"
+Sphinx = "^8.0.0"
 # Sphinx - Additional modules
-myst-parser = "^2.0.0"
+myst-parser = "^4.0.0"
 sphinx-book-theme = "^1.0.1"
 sphinx-favicon = "^0.2"
 sphinxcontrib-contentui = "^0.2.5"
-sphinxcontrib-images = "^0.9.4"
+sphinxcontrib-images = "^1.0.1"
 sphinx-icon = "^0.1.2"
 sphinx-autobuild = "^2021.3.14"
 sphinx-autoapi = "^3.0.0"
```

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
+
+You can also use other linked data vocabularies. To do this, you need to identify them with a prefix and register them
+with the data model by passing it `extra_vocabs` as a `dict` mapping prefixes to URLs where the vocabularies are
+provided as JSON-LD:
+
+```{code-block} python
+:caption: Injecting additional schemas
+from hermes.model import SoftwareMetadata
+
+# Contents served at https://bar.net/schema.jsonld:
```
zyzzyxdonta: Please only use reserved example domains! https://en.wikipedia.org/wiki/Example.com

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
+
+You need to be able to access data in the data model instance to add, edit or remove data.
+Data can be accessed by using term strings, similar to how values in Python `dict`s are accessed by keys.
+
+```{important}
+When you access data from a data model instance,
+it will always be returned in a **list**-like object!
+```
+
+The reason for providing data in list-like objects is that JSON-LD treats all property values as arrays.
```
zyzzyxdonta: Link to the related section in the JSON-LD spec would be nice

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
+# $> Author has only emails with their name in it.
+```
+
+The example continues to show how to assert data values.
+
+As mentioned in the [introduction to the data model](#data-model), 
+`hermes` uses a JSON-LD-like internal data model. 
+The API class {class}`hermes.model.SoftwareMetadata` hides many
+of the more complex aspects of JSON-LD and makes it easy to work
+with the data model.
```
zyzzyxdonta: This feels like a misplaced summary. I would just remove it.

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
+
+So the API class hides the internal model objects.
+Therefore, they work as you would expect from plain
+Python data:
+
+```{code-block} python
+:caption: Naive containment assertion that raises
```
zyzzyxdonta: It doesn't though. The mock output is from the path where the assertion holds.

docs/source/dev/data_model.md
```diff
@@ -1,27 +1,286 @@
...
+#            }]
+#        }
+#    )
+```
+
+---
+
+## See Also
+
+- API reference: {class}`hermes.model.SoftwareMetadata`
```
zyzzyxdonta: This link never works throughout the whole docs. For some reason, the class is only exposed in the docs as `hermes.model.api.SoftwareMetadata`

docs/source/dev/plugin_structure.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
 # Write a plugin for HERMES
  
 
-This tutorial will present the basic steps for writing an additional harvester.
-At the moment only the architecture for harvester plugins is stable.
-The full code and structure is available at  [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
+This tutorial will present the basic steps for writing additional plugins.
+
+The full code and structure of a harvest plugin is available at [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
```
zyzzyxdonta, line 20: Emphasize that this is _an example_ of a "standalone" (as in out-of-hermes-source) harvest plugin.

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
...
  
 
-This tutorial will present the basic steps for writing an additional harvester.
-At the moment only the architecture for harvester plugins is stable.
-The full code and structure is available at  [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
+This tutorial will present the basic steps for writing additional plugins.
+
+The full code and structure of a harvest plugin is available at [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
 This plugin extracts information from the local git history.
 The hermes-plugin-git will help to gather contributing and branch metadata.
```
zyzzyxdonta:
```suggestion
The hermes-plugin-git will help to gather contribution and branch metadata.
```

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
...
+
 ```{note}
 For this tutorial you should be familiar with HERMES.
```
zyzzyxdonta:
```suggestion
To follow this tutorial you should be familiar with HERMES. 
```

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
...
+The full code and structure of a harvest plugin is available at [hermes-plugin-git](https://github.com/softwarepub/hermes-plugin-git).
 This plugin extracts information from the local git history.
 The hermes-plugin-git will help to gather contributing and branch metadata.
+
 ```{note}
 For this tutorial you should be familiar with HERMES. 
-If you never used HERMES before, you might want to check the tutorial: [Automated Publication with HERMES](https://docs.software-metadata.pub/en/latest/tutorials/automated-publication-with-ci.html).
+If you never used HERMES before, you might want to check the tutorial: [Automated Publication with HERMES](./automated-publication-with-ci).
+
+Also all metadata directly handled by HERMES is [JSON-LD](https://json-ld.org/) so you should be familiar with that when writing a plugin.
```
zyzzyxdonta: Link to the "JSON-LD for plugin developers" guide instead.

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
...
+And uses the [schmea.org](https://schema.org/) (with prefix "schema") and the [CodeMeta](https://codemeta.github.io/) (without prefix) context.
```
zyzzyxdonta:
```suggestion
And uses the [schema.org](https://schema.org/) (with prefix "schema") and the [CodeMeta](https://codemeta.github.io/) (without prefix) context.
```

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -14,26 +14,35 @@ SPDX-FileContributor: Oliver Bertuch
...
 The structure for every plugin follows the same schema.
-There is a top-level base class for every plugin. In this `HermesPlugin` class there is one abstract method `__call__` which needs to be overwritten.
-Furthermore, the `HermesCommand` class provides all needs for writing a plugin used in a HERMES command.
-So the `HermesPlugin`s call method gets an instance of the `HermesCommand` that triggered this plugin to run.
-In our case this will be the `HermesHarvestCommand` which calls all harvest plugins.
-The plugin class also uses a derivative of `HermesSettings` to add parameters that can be adapted by the configuration file.
-`HermesSettings` are the base class for command specific settings.
+Every plugin is a sub class of a sub class of the {py:class}`~hermes.commands.base.HermesPlugin` class.
+This class implements one abstract method, {py:meth}`~hermes.commands.base.HermesPlugin.__call__`, which needs to be overwritten by every plugin.
+In between the {py:class}`~hermes.commands.base.HermesPlugin` class and the class of a specific plugin there is another class which follows the naming scheme `Hermes{Step}Plugin` where `{Step}` is the step the plugin is for.
```
zyzzyxdonta:
```suggestion
In between the {py:class}`~hermes.commands.base.HermesPlugin` class and the class of a specific plugin there is another class which follows the naming scheme `Hermes{Step}Plugin` where `{Step}` is the workflow step the plugin is associated with.
```

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
-In the `hermes.toml` this would be achieved by [harvest.{plugin_name}].
-The `GitHarvestSettings` are associated with the `GitHarvestPlugin`.
-In the plugin you need to overwrite the `__call__` method.
-For now a simple "Hello World" will do. The method returns two dictionaries.
-These will contain the harvested data in CodeMeta (JSON-LD) and additional information, e.g., to provide provenance information.
-That is the basic structure for the plugins source code.
-
-To integrate this code, you have to register it as a plugin in the `pyproject.toml`.
+
+The {py:meth}`~hermes.commands.harvest.base.HermesHarvestPlugin.__call__` method of harest plugins needs to return a {py:class}`~hermes.model.api.SoftwareMetadata` object containing the harvested metadata.
```
zyzzyxdonta:
```suggestion
The {py:meth}`~hermes.commands.harvest.base.HermesHarvestPlugin.__call__` method of harvest plugins needs to return a {py:class}`~hermes.model.api.SoftwareMetadata` object containing the harvested metadata.
```

docs/source/dev/plugin_structure.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
+
+class YourProcessSettings(BaseModel):
+    # TODO: add your settings
+    pass
+
+
+class YourProcessPlugin(HermesProcessPlugin):
+    settings_class = YourProcessSettings
+
+    def __call__(self, command: HermesProcessCommand) -> dict[Union[str, None], dict[Union[str, None], MergeAction]]:
```
zyzzyxdonta: I think it would be nice to have type aliases here to make it more understandable what these mappings are.

zyzzyxdonta: Mapping from `None` to something feels weird. There can only be one `None` key. What would this be used for?

docs/source/dev/plugin_structure.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
+        strategies = {}
+
+        # TODO: define the merge strategies that will be used by HERMES
+
+        return strategies
+```
+
+The {py:meth}`~hermes.commands.process.base.HermesProcessPlugin.__call__` method of process plugins needs to return a dictionary mappings strings and/ or `None` to dictionaries mapping strings or `None` to {py:class}`~hermes.model.merge.action.MergeAction`.
```
zyzzyxdonta: This needs to be explained better. The sentence only repeats what is already in the type annotation. What do I map? I.e. what does the key of type `str` represent? What does a key `None` represent? What's the first layer of the mapping, what is the second layer?

docs/source/dev/plugin_structure.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
+strategies = {
+    full_type_iri: {
+        full_property_iri: Reject(),
+        ...
+    },
+    ...
+}
+```
+
+HERMES would use the {py:class}`~hermes.model.merge.action.Reject` strategy for merging values of the key `full_property_iri` in objects of type `full_type_iri`. (A key in strategies being `None` instead of a string indicates to HERMES that its value is to be used as a default [i.e. if no more specific entry exists].)
```
zyzzyxdonta, line 134: This feels hacky. Have you considered [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)?

docs/source/dev/plugin_structure.md
zyzzyxdonta, line 1: This whole file feels more like a reference than a tutorial. A tutorial should be "learning by doing". An example could be having the reader re-build the git-plugin from scratch.

See also: https://diataxis.fr/tutorials/

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
+The settings of your plugin can be accessed via `self.settings.{plugin_step}.{plugin_name}.{attribute_name}`.
+And setting it in the `hermes.toml` works like this:
+```shell
+[{plugin_step}.{plugin_name}]
+{attribute_name} = value
+```
+
+## Configure HERMES to use your plugin
+
+To integrate your plugin, you have to register it as a plugin in the `pyproject.toml`.
```
zyzzyxdonta: I would prefer calling the thing in the `pyproject.toml` "entry point". That's also how it's called in the PEP: https://peps.python.org/pep-0621/#entry-points


```suggestion
To integrate your plugin, you have to register it as an entry point in the `pyproject.toml`.
```

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -42,42 +51,233 @@ E.g., you can set your authentication token for InvenioRDM by adding the followi
...
 To learn more about the `pyproject.toml` check https://python-poetry.org/docs/pyproject/ or refer to [PEP621](https://peps.python.org/pep-0621/).
```
zyzzyxdonta: I would remove the Poetry mention. Instead we can link to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ which gives an agnostic overview.

docs/source/tutorials/writing-a-plugin-for-hermes.md
```diff
@@ -113,59 +313,76 @@ This variant is used to contribute to the HERMES community or adapt the HERMES w
...
 Note that this differs with the accessibility and your wishes, check [Explicit Package Sources](https://python-poetry.org/docs/repositories/#explicit-package-sources).
 
 The second thing to adapt is to declare the access point for the plugin.
-You can do that with `git = "hermes_plugin_git.harvest:GitHarvestPlugin"`.
-This expression makes the `GitHarvestPlugin` from the `hermes_plugin_git` package, a `hermes.harvest` plugin named `git`.
+You can do that with `{plugin_name} = "{plugin_package}.{plugin_module}:{plugin_class}"`.
+This expression makes the `plugin_class` from the `plugin_package` package, a `hermes.{plugin_step}` plugin named `plugin_name`.
 So you need to configure this line with your plugin properties.
 
 Now you just need to add the plugin to the `hermes.toml` and reinstall the adapted poetry package.
```
zyzzyxdonta: Same here. Poetry mentions should be removed. Instead we can use the generic version `[project.entry-points."hermes..."]` now. All of the Poetry stuff was written here because `pyproject.toml` support was poor in other tools years ago, when this was written.

docs/source/tutorials/writing-a-plugin-for-hermes.md
zyzzyxdonta: As https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ also uses the word plugin, I think it would be best to also mention it here, but then use the term "entry point" for consistency and to differentiate the HERMES plugin and the line in the `pyproject.toml`.

pyproject.toml
notactuallyfinn: I removed `toml` from the list of dependencies.
`tomli` can't be removed and does not support write operations (and therefore can't replace the other toml dependencies.)

pyproject.toml
notactuallyfinn: I renamed `DoNothingCuratePlugin` to `PassCuratePlugin` (and everything related) for now, because `passCurate` is shorther than `DoNothingCurate`.

poetry.lock
notactuallyfinn: I had to remove 'sphinxemoji' too or constrain `setuptools` to <82 (the dependency `pkg_resources` (deprecated) was removed in `setuptools` version 82 but was necessary for 'sphinxemoji').

Also building the docs loally always (at least for me) results in errors from myst (failed linkings).
This has been an issue for longer though, but I haven't been able to fix it so far.

docs/source/tutorials/writing-a-plugin-for-hermes.md
notactuallyfinn: Is the way I rephrased it fine with you?

poetry.lock
zyzzyxdonta: Yup, the setuptools change was exactly the same problem I had. I didn't have any problems with sphinxemoji though. Strange ... 🤔 

Yes, the errors from myst have been there for a while. I think it's mostly unused or missing references/labels. I haven't looked into it.

docs/source/tutorials/writing-a-plugin-for-hermes.md
zyzzyxdonta: Very good! 👍🏻

src/hermes/model/hermes_cache.py
notactuallyfinn: I renamed `context_manager` to `hermes_cache`, `HermesContext` to `HermesCacheManager` and `HermesContextError` to `HermesCacheError`. (`HermesCache` was not renamed.)
Do those names seem better/ good enough?

zyzzyxdonta: Yes, a lot better! I'm not convinced we need `Hermes` in the names but I can accept it :-)

src/hermes/model/types/ld_container.py
notactuallyfinn: `ld_container` implements `to_python` and I didn't know what to call this version of an `ld_container`.

src/hermes/model/types/pyld_util.py
notactuallyfinn: I only remember that there was some bug with the compaction. This fix was written by @SKernchen, who might be able to answer your questions.

As for the function of `"none"`: I believe it tells the JSON-LD processor that a value should be dropped.

poetry.lock
notactuallyfinn: Removing 'sphinxemoji' caused no errors or warnings. Therefore leaving it removed should be fine.

I'll look into it when I find time.

src/hermes/model/types/ld_container.py
zyzzyxdonta: How about calling it `PYTHON_COMPATIBLE_LD_CONTAINER` and adding an explanation to the docstring? Or `NATIVE_LD_CONTAINER`?

docs/source/dev/plugin_structure.md
notactuallyfinn: `None` tells hermes that the value is supposed to be a default value, i.e. if no key fits.

zyzzyxdonta: Yup, I figured this out later. See also my other comment below. Sorry, forgot to delete this one 😅

docs/source/dev/plugin_structure.md
notactuallyfinn: No, I hadn't.

But now that I looked into it, I don't see the benefit.
Sure the `None` keys would be no longer necessary, which means in one spot no checking for none existing keys. Instead we need to handle the `default_factory` while merging the strategies.
(And I find it more intuitive with `None`.)

docs/source/dev/plugin_structure.md
notactuallyfinn: Only here or also in the source code?

docs/source/dev/plugin_structure.md
notactuallyfinn: I added a few sentences to keep the links and add some more clarity.
How do you like it now?

docs/source/dev/plugin_structure.md
notactuallyfinn: I moved it into the `devs` section and will add real tutorials, but that may take some time.

Should there be one tutorial for each steps plugin, or could one be enough?

docs/source/dev/data_model.md
notactuallyfinn: I'll look through the entire file and correct all other outdated information.

docs/source/dev/plugin_structure.md
zyzzyxdonta: OK

docs/source/dev/plugin_structure.md
zyzzyxdonta: No problem. We can add a tutorial after this is merged.

I think one tutorial would be enough as long as it covers the most important points. To me that would be:

- How to get HERMES to find the plugin
- How to implement configuration via hermes.toml
- How to interact with the data model
- How to read/write the data out of/into the cache

docs/source/dev/plugin_structure.md
zyzzyxdonta: Also in the code.

How about something like this:

```python
IRI: TypeAlias = str
ObjectTypeIRI: TypeAlias = IRI
PropertyIRI: TypeAlias = IRI

PropertyStrategies: TypeAlias = Mapping[PropertyIRI, Strategy]  # or whatever the Strategy base class is called
ObjectTypeStrategies: TypeAlias = Mapping[ObjectTypeIRI, PropertyStrategies]

class YourProcessPlugin(HermesProcessPlugin):
    
    ...
    
    def __call__(self, command: HermesProcessCommand) -> ObjectTypeStrategies:
        ...
```

docs/source/dev/plugin_structure.md
zyzzyxdonta: I don't see it 🤔

src/hermes/model/types/pyld_util.py
SKernchen: Yes the pyld-lib could not find the right compacted iri for
`"author": { "@id": "schema:author", "@container": "@list" },`
(I guess it is an error from them). Because of the @list in @container
So I manually corrected it.
But yes, I'll write a comment for that

poetry.lock
notactuallyfinn: From what I found looking into the myst errors is that the problem is that we use two symlinks (to LICENSES and adr).
At least on my Windows machine they get converted into plain text files leading to problems with the myst links.

I'd suggest that we, at least for now, use absolut or relative paths without symlinks.
After trying it out:
Removing the symlink to LICENSES was fine (paths hat do be changed).
Removing the symlink to adr wasn't easy. The only option I found is to move the subfoulder containing the ADRs. (toctrees can't contain references outside of the project folder.)

src/hermes/model/types/pyld_util.py
zyzzyxdonta: Huh, I see what you mean. This seems to be specific to `@list`. With `@set` it works 🤔 

```python
#!/usr/bin/env -S pipx run
# /// script
# requires-python = ">=3.10"
# dependencies = [
#   "PyLD (>=3.1.0)",
# ]
# ///

import json

from pyld import jsonld


doc_str = """
[
  {
    "@id": "_:Nc1239476e3cc45e387cfa0cf9b519b91",
    "@type": [
      "https://schema.org/SoftwareApplication"
    ],
    "https://schema.org/author": [
      {
        "@id": "_:Nd3abed4ba38842d8aaf46464efef5090"
      }
    ]
  },
  {
    "@id": "_:Nd3abed4ba38842d8aaf46464efef5090",
    "@type": [
      "https://schema.org/Person"
    ],
    "https://schema.org/givenName": [
      {
        "@value": "John"
      }
    ],
    "https://schema.org/lastName": [
      {
        "@value": "Doe"
      }
    ]
  }
]
"""

doc = json.loads(doc_str)

context = {
    # "author": { "@id": "https://schema.org/author" },  # works
    # "author": {"@id": "https://schema.org/author", "@container": "@set"},  # also works
    "author": {"@id": "https://schema.org/author", "@container": "@list"},  # doesn't work
}

compacted = jsonld.compact(doc, context)
compacted_text = json.dumps(compacted, indent=2)
print(compacted_text)

assert '"https://schema.org/author":' not in compacted_text
```

This yields:

```json
{
  "@context": {
    "author": {
      "@id": "https://schema.org/author",
      "@container": "@list"
    }
  },
  "@graph": [
    {
      "@id": "_:Nc1239476e3cc45e387cfa0cf9b519b91",
      "@type": "https://schema.org/SoftwareApplication",
      "https://schema.org/author": {
        "@id": "_:Nd3abed4ba38842d8aaf46464efef5090"
      }
    },
    {
      "@id": "_:Nd3abed4ba38842d8aaf46464efef5090",
      "@type": "https://schema.org/Person",
      "https://schema.org/givenName": "John",
      "https://schema.org/lastName": "Doe"
    }
  ]
}
```
```text
Traceback (most recent call last):
  File "/home/pape58/pyld-test.py", line 58, in <module>
    assert '"https://schema.org/author":' not in compacted_text
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

Maybe consider raising an issue with pyld 🤷🏻‍♂️

SKernchen: Thanks, I did
https://github.com/digitalbazaar/pyld/issues/292

zyzzyxdonta: OK, as discussed in that issue, the bug is on our side. The data should look like this (i.e. with `@list` for the authors):

```json
[
  {
    "@id": "_:Nc1239476e3cc45e387cfa0cf9b519b91",
    "@type": [
      "https://schema.org/SoftwareApplication"
    ],
    "https://schema.org/author": { 
       "@list" : [
          {
            "@id": "_:Nd3abed4ba38842d8aaf46464efef5090"
          }
        ]
    }
  },
  {
    "@id": "_:Nd3abed4ba38842d8aaf46464efef5090",
    "@type": [
      "https://schema.org/Person"
    ],
    "https://schema.org/givenName": [
      {
        "@value": "John"
      }
    ],
    "https://schema.org/lastName": [
      {
        "@value": "Doe"
      }
    ]
  }
]
```

I guess it must be implemented in the data model then, that for known contexts, the correct container type must be applied.

zyzzyxdonta: Do you think these would be reasonable test cases?

```python
def test_container_type_from_single():
    data = SoftwareMetadata()
    data["schema:author"] = {"@type": "schema:Person", "schema:name": "John Doe"}
    assert data.ld_value == [
        {
            "http://schema.org/author": {
                "@list": [
                    {
                        "@type": ["http://schema.org/Person"],
                        "http://schema.org/name": [{"@value": "John Doe"}],
                    }
                ]
            }
        }
    ]


def test_container_type_from_list():
    data = SoftwareMetadata()
    data["schema:author"] = [
        {"@type": "schema:Person", "schema:name": "John Doe"},
        {"@type": "schema:Person", "schema:name": "Jane Doe"},
    ]
    assert data.ld_value == [
        {
            "http://schema.org/author": {
                "@list": [
                    {
                        "@type": ["http://schema.org/Person"],
                        "http://schema.org/name": [{"@value": "John Doe"}],
                    },
                    {
                        "@type": ["http://schema.org/Person"],
                        "http://schema.org/name": [{"@value": "Jane Doe"}],
                    },
                ]
            }
        }
    ]
```

notactuallyfinn: While playing around with this to understand what was working how, I found this:
Compacting the following expanded JSON-LD value results in two different objects when compacting with `pyld.jsonld.compact` and via the Playground:
```
[{
    "http://schema.org/author": [{
        "@list": [{
            "@type": ["http://schema.org/Person"],
            "http://schema.org/name": [{"@value": "Testi"}]
          }]
      }]
}]
```
```
{"@context": ["https://doi.org/10.5063/schema/codemeta-2.0", {"schema": "http://schema.org/"}]}
```
- [Playground](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=%5B%7B%22http%3A%2F%2Fschema.org%2Fauthor%22%3A%5B%7B%22%40list%22%3A%5B%7B%22%40type%22%3A%5B%22http%3A%2F%2Fschema.org%2FPerson%22%5D%2C%22http%3A%2F%2Fschema.org%2Fname%22%3A%5B%7B%22%40value%22%3A%22Testi%22%7D%5D%7D%5D%7D%5D%7D%5D&context=%7B%22%40context%22%3A%5B%22https%3A%2F%2Fdoi.org%2F10.5063%2Fschema%2Fcodemeta-2.0%22%2C%7B%22schema%22%3A%22http%3A%2F%2Fschema.org%2F%22%7D%5D%7D) preserves the `@list`.
- `pyld` removes it:
```
import json
from pyld import jsonld

doc_str = """
[{"http://schema.org/author": [{"@list": [{"@type": ["http://schema.org/Person"], "http://schema.org/name": [{"@value": "Testi"}]}]}]}]
"""

doc = json.loads(doc_str)

context = ["https://doi.org/10.5063/schema/codemeta-2.0", {"schema": "http://schema.org/"}]

compacted = jsonld.compact(doc, context)
compacted_text = json.dumps(compacted, indent=2)
print(compacted_text)
```

I'm not sure which version is wrong but both can't be right, right?

zyzzyxdonta: You have

```
"http://schema.org/author": [{"@list":
```

Shouldn't that be

```
"http://schema.org/author": {"@list":
```

?

(sure, this doesn't explain the difference in outputs)

zyzzyxdonta: OK, no. This doesn't change anything.

zyzzyxdonta: No wait. I can't reproduce this. If I take your code from above:

```python
import json
from pyld import jsonld

doc_str = """
[{"http://schema.org/author": [{"@list": [{"@type": ["http://schema.org/Person"], "http://schema.org/name": [{"@value": "Testi"}]}]}]}]
"""

doc = json.loads(doc_str)

context = ["https://doi.org/10.5063/schema/codemeta-2.0", {"schema": "http://schema.org/"}]

compacted = jsonld.compact(doc, context)
compacted_text = json.dumps(compacted, indent=2)
print(compacted_text)
```

And I ran it in a brand new Python 3.14 environment with PyLD 3.1.0 and Requests installed, I get exactly what the playground produces, including the `@list` annotation:

```json
{
  "@context": [
    "https://doi.org/10.5063/schema/codemeta-2.0",
    {
      "schema": "http://schema.org/"
    }
  ],
  "author": {
    "@list": [
      {
        "type": "Person",
        "name": "Testi"
      }
    ]
  }
}
```

So the problem must be in the cached documents loaded by hermes.

SKernchen: So the first test works with all three with http:

```
assert '"http://schema.org/author":' not in compacted_text
```

as assertion, it works no matter which one is used in @id .
Only with @list and `assert https` it does not work if @id is https.

SKernchen: Ok so similar stuff with the last reproduction attempt
```
context = ["http://doi.org/10.5063/schema/codemeta-2.0", {"schema": "http://schema.org/"}] # without @list
context = ["https://doi.org/10.5063/schema/codemeta-2.0", {"schema": "http://schema.org/"}] # with @list
context = [{"schema": "http://schema.org/"}] # with @list
```
So the times it works is when the codemeta schema is not found ^^

SKernchen:
Another thing I noticed is, that with pyld 3.0 some parts are not working anymore. I made an issue for this. See #483

notactuallyfinn (approved):

PR #488, opened 2026-08-18
Use HERMES' Zenodo concept DOI for "isCompiledBy" relations

### LINKED ISSUE #487: Keep Hermes DOI used for `IsCompiledBy` relation up to date
When making a deposit to an Invenio- or InvenioRDM-based platform, hermes sets a `IsCompiledBy` relation on the created record, to indicate that it was created by hermes. Currently, the DOI used for this, is hard-coded.

https://github.com/softwarepub/hermes/blob/58c642c348a13b1f8527d44c8cf74c00a9e13a42/src/hermes/commands/deposit/invenio.py#L388

This leads to two issues:

1. The DOI needs to be updated manually which can easily be forgotten.
2. We run into a chicken-and-egg problem. While a new version of hermes is being published and pushed to PyPI, the DOI of this version is not known yet and can't be included in the release.

There are two different ways to fix this:

1. Use the hermes concept DOI as the related identifier
    - ➕ Easy to implement
    - ➖ We "loose track" of which hermes versions are used and we loose valuable debugging information in case we find issues in publications. Though this might not be too bad as hermes is still in its infancy.
2. Figure out the DOI at runtime
    - ➕ Enables more insights
    - ➖ Complex implementation that needs to find all possible hermes DOIs on DataCite and then select the one which matches best (could go wrong with pre-released versions, forks, in development, etc.)

zyzzyxdonta:
When depositing to Invenio- and InvenioRDM-based platforms, HERMES adds a "isCompiledBy" relation to the deposit metadata.

This pull request updates the code to use HERMES' concept DOI (`10.5281/zenodo.13221383`) instead of the frequently out-of-date "current" DOI. While this makes the relation metadata slightly less helpful (e.g. we can't track any more which HERMES version introduced a potential issue in metadata generation), at least we don't use wrong version metadata for the depositions.

Closes #487

This pull request also updates the "current" HERMES DOI to the latest published version 0.9.1.

### DIFF
--- a/src/hermes/commands/deposit/invenio.py
+++ b/src/hermes/commands/deposit/invenio.py
@@ -22,7 +22,7 @@ from hermes.commands.deposit.error import DepositionUnauthorizedError
 from hermes.error import MisconfigurationError
 from hermes.model.context import CodeMetaContext
 from hermes.model.path import ContextPath
-from hermes.utils import hermes_doi, hermes_user_agent
+from hermes.utils import hermes_concept_doi, hermes_user_agent
 
 
 _log = logging.getLogger("cli.deposit.invenio")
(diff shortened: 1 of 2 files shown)


SKernchen (approved):
Havent we not published 0.10.0 already? I do not see a record but i thought we did.

PR #492, opened 2026-09-03
Adapt `file_exists` harvester to the new data model

zyzzyxdonta:


SKernchen (approved):
We act on codemeta:readme later

